PR #35, opened 2023-10-26
Feature t2k support

EdAtkin:
This is now ready to be approved!

- covarianceXsec can now be initialised from a yaml file and parse all the information through from this
- there was a small tidy up of covarianceBase and some of the member variable naming schemes to be clearer
- covarianceBase now support proposing from a different matrix to the prior covariance and should be able to do adaptive step-size tuning but not validated
- Now can use CUDAProb for beam, and the OscClass for atmospheric calculations like has been done on DUNE for a while.
- spline monolith structure in the splineFDBase object. This makes the spline evaluation faster and is straight from T2K
- Functional parameters now can be called without a problem.
- I have been using this branch to validate the T2K version of the refactor and see perfect agreement in LLH scans

Other things to be done:
- clean up function names, getters and setters in covarianceBase further
- fix a small bug in the MC stat calculation which I saw in T2K. Not sure of the problem but probably a simple fix.
- general code tidy-up as there are several commented out patches still
- check that GPU support is fine and also that all compilation options work (use prob3)
- check various different compilation flags
- documentation!


samplePDF/samplePDFFDBase.h
```diff
@@ -210,30 +226,21 @@ class samplePDFFDBase : virtual public samplePDFBase , virtual public interfaceP
...
   // An axis to set binned oscillation weights
   TAxis *osc_binned_axis ;
   //===============================================================================
   
-  //Variables controlling nuebar appearance
-  double Beta;
-  bool useBeta;
-  bool applyBetaNue;
-  bool applyBetaDiag;
-
```
KSkwarczynski, line 225: good at cleaning this, no idea why it was here in the first place...

samplePDF/samplePDFFDBase.cpp
```diff
@@ -13,24 +13,15 @@ samplePDFFDBase::samplePDFFDBase(double pot, std::string mc_version, covarianceX
...
-
   samplePDFFD_array = NULL;
   samplePDFFD_data = NULL;
   
   //KS: For now FD support only one sample
   nSamples = 1;
   SampleName.push_back("FDsample");
+
+  //ETA - make sure osc is NULL!!
+  //Osc = NULL;
```
KSkwarczynski, line 24: why is  this commented?

samplePDF/Structs.h
```diff
@@ -1044,4 +1078,64 @@ double returnCherenkovThresholdMomentum(int PDG);
...
+  case 1:
+    switch (abs(nu_f)) {
+    case 1:
+      return CUDAProb_nu::e_e;
+      break;
+    case 2:
+      return CUDAProb_nu::e_m;
+      break;
+    case 3:
+      return CUDAProb_nu::e_t;
```
KSkwarczynski, line 1111: I think these should be in CUDAprob stuct or something. It isn't intuitive that if you want to find some CUDAprob specyfic stuff you go to struct

mcmc/mcmc.cpp
```diff
@@ -253,74 +250,9 @@ void mcmc::ProposeStep() {
...
-      std::vector<double> oscpars;
-      std::vector<double> oscpars2;
-
-      for (int par = 0; par < int(osc->getSize()+3); par++) {
-        // skip the extra entries
-        if (par==6 || par==7 || par==8) {
-          continue;
-
-          // normal osc pars
-        } else if (par<6) {
```
KSkwarczynski, line 281: Thank you for clearing this clussterfuck of a mess

manager/manager.h
```diff
@@ -48,6 +40,8 @@ class manager {
   YAML::Node config;
 
 /*
+ * Keeping this all here as we should start adding in functions here to do reading
```
KSkwarczynski, line 43: For T2K we really need save setting
https://github.com/t2k-software/MaCh3/blob/6b620ab3b38b843ac269f7ee7e9c7688f7adb6e6/manager/manager.cpp#L1730

Maybe not so huge but we need stuff like location of XsecMatrix, Osc matrix etc

covariance/covarianceXsec.h
```diff
@@ -237,6 +264,16 @@ class covarianceXsec : public covarianceBase {
...
+  private:
+	//std::vector<std::string> _fNames;
+	std::vector<std::string> _fFancyNames;
+
+	//ETA - do we need these now?
+	std::vector<std::vector<int>> _fNormModes;
+	std::vector<std::string> _fNDSplineNames;
+	std::vector<std::string> _fFDSplineNames;
+	std::vector<std::vector<int>> _fFDSplineModes;
+	int _nNormPars;
```
KSkwarczynski, line 276: In long term I agree we don't need separate ND and FD spline names

covariance/covarianceXsec.h
```diff
@@ -171,6 +195,7 @@ class covarianceXsec : public covarianceBase {
     double *xsec_param_prior_a;
 
     // Contains the parameter names
+	// ETA - don't think we need this anymore tbh
     std::vector<std::string> xsec_param_names;
```
KSkwarczynski, line 199: We should use parameters name from CovBase

covariance/covarianceXsec.h
```diff
@@ -12,42 +12,52 @@
...
-    int    GetNumParams()               {return nPars;}
-    const char* GetParName(int i) const {return xsec_param_names[i].c_str();}
-
-    bool isParFlux(int i){
+	// ETA - a lot of these can go... they're just duplications from the base
+	// class.
+    const double GetParamUpperBound(const int i) {return xsec_param_ub_a[i];}
+    const double GetParamLowerBound(const int i) {return xsec_param_lb_a[i];}
+    const double GetParamPrior(const int i)      {return xsec_param_prior_a[i];}
+    const int  GetXSecParamID(const int i, const int j) const {return xsec_param_id_a[i][j];}
```
KSkwarczynski, line 34: Agreed and we can put inline ;)

covariance/covarianceXsec.h
```diff
@@ -12,42 +12,52 @@
...
-#include "throwParms/ThrowParms.h"
+
+#include "yaml-cpp/yaml.h"
 
 class covarianceXsec : public covarianceBase {
 
   public:
-  covarianceXsec(const char *name, const char *file, double threshold=-1,int firstpcapar=-999,int lastpcapar=-999);
+  covarianceXsec(const char *name, const char *file, double threshold=-1,int FirstPCAdpar=-999,int LastPCAdpar=-999);
+  covarianceXsec(const char *YAMLFile);
```
KSkwarczynski, line 22: Why there is no Yaml in PCA consturctor?

covariance/covarianceXsec.cpp
```diff
@@ -94,59 +112,321 @@ covarianceXsec::covarianceXsec(const char *name, const char *file, double thresh
...
+	//Get the map of parameter name to correlation fomr the Correlations object
+	for (auto const& [key, val] : Correlations[i]) {
+	  int index = -1;
+
+	  //If you found the parameter name then get the index
+	  if (CorrNamesMap.find(key) != CorrNamesMap.end()) {
+		index=CorrNamesMap[key];
+	  }
+	  else {
+		std::cout << "Parameter " << key << " not in list! Check your spelling?" << std::endl;
```
KSkwarczynski, line 386: We convert Yaml to ROOT matrix and then in CovBase ROOT to double** ;)
Which is fine for now

covariance/covarianceOsc.h
```diff
@@ -23,7 +20,15 @@ class covarianceOsc : public covarianceBase
...
       void setExtraBranches(TTree &tree);
-      
+
+      double GetPathLength() {
+        return L;
+      }
+
+      double GetDensity() {
+        return density;
+      }
```
KSkwarczynski, line 30: In long term we should make them proper syst parameters. In short term we can inline ;)

KSkwarczynski (commented):
Glad to see, have minor comments for now. Might give it another real later.

samplePDF/CMakeLists.txt
```diff
@@ -2,27 +2,33 @@ set(SamplePDF_implementation_files
...
 set(SamplePDF_header_files 
 	samplePDFBase.h
 	samplePDFFDBase.h
 	interfacePDFEbE.h
 	Structs.h )
 
-add_library(SamplePDF SHARED ${SamplePDF_implementation_files})
+if(NOT ${CPU_ONLY})
+  list(APPEND SamplePDF_implementation_files probGpu.cu)
+endif()
```
KSkwarczynski, line 15: This is bit messy but you right now you compile probGpu even if using CUDAprob. I know this should be fixed with Dan't oscialtor but all this stuff should be moved from SamplePDF

covariance/covarianceXsec.h
pgranger23: Just a small comment on all these `const double/int ...` functions. I don't know what your intent is there with the `const` qualifier, but it currently does not mean anything to the compiler. If the intent is to say that the function does not affect the class object, `const` should be moved at the end of the signature. Otherwise, as you are returning by value I'm pretty sure it has no impact/meaning.

covariance/covarianceBase.h
```diff
@@ -34,96 +39,139 @@ class covarianceBase {
...
-  const std::vector<double> & getNominalArray() const { return nominal; };
+  virtual std::vector<double> getNominalArray();
   const std::vector<double> getProposed() const;
-  double getNominal(int i) { 
-    return nominal[i]; };
-  double getParProp(int i) { 
-    return fParProp[i]; 
+  const double getParProp(const int i) {
+    return _fPropVal[i]; 
+  };
```
KSkwarczynski, line 147: Could you check if Henry's updated to covariance in this PR 
https://github.com/t2k-software/MaCh3/pull/548/files
are included, otherwise something may not work with T2K

covariance/covarianceBase.h
```diff
@@ -34,96 +39,139 @@ class covarianceBase {
...
+  const double getParInit(const int i) {
+    return _fPreFitValue[i];
   };
-  double getParInit(int i) { 
-    return fParInit[i];
+
+  virtual const double getNominal(const int i) {
+    return getParInit(i);
   };
-  const double GetLowerBound(const int i) { return fParLoLimit[i];};
```
henry-israel, line 117: Really minor comment, a lot of these methods don't need to be "const" since the compiler already identifies this

covariance/covarianceBase.h
```diff
@@ -34,96 +39,139 @@ class covarianceBase {
...
-    return fParProp[i]; 
+  const double getParProp(const int i) {
+    return _fPropVal[i]; 
+  };
+  const double getParCurr(const int i) {
+    return _fCurrVal[i];
   };
-  double getParCurr(int i) { 
-    return fParCurr[i];
+  const double getParInit(const int i) {
```
henry-israel, line 151: minor comment, this doesn't need to be const [the compiler ignores it + it flashes up a load of warnings at compile time!]

covariance/covarianceXsec.cpp
```diff
@@ -571,133 +746,179 @@ void covarianceXsec::scanParameters() {
...
+
+		  //ETA - This can be a vector :) can provide different kinematic variables to cut on
+		  tmp_xsec.KinematicVarStr = _fKinematicPars[i];
+
+		  //ETA - This can be a vector :) can provide different kinematic variables to cut on
+		  std::vector< std::vector<double> > Selections(_fKinematicPars[i].size());
+
+		  //ETA - push back kinematic type with dummy -999 since this needs to be converted into an enum for a kinematic type within
+		  //a samplePDFFD daughter class
+		  for(int KinVar_i = 0 ; KinVar_i < _fKinematicPars[i].size() ; ++KinVar_i) {
```
henry-israel, line 771: Super nitpicky but can you cast _fKinematicPars[i].size() to an integer (or KinVar_i to size_t)

samplePDF/Structs.h
henry-israel: Also it currently doesn't have a default case so this might cause issues

samplePDF/samplePDFFDBase.cpp
```diff
@@ -788,6 +863,132 @@ void samplePDFFDBase::CalcXsecNormsBins(int iSample){
...
+  MCSamples[iSample].Oscillator->setDensity(Density);
+#else
+//if we're not mulithreading then just set it to 1
+    MCSamples[iSample].Oscillator = new cudaprob3::BeamCpuPropagator<double>(MCSamples[iSample].nEvents, 1);
+  MCSamples[iSample].Oscillator->setPathLength(PathLength);
+  MCSamples[iSample].Oscillator->setDensity(Density);
+#endif //MULTITHREAD
+#else
+//if we want to use CUDAProb3 GPU then setup BeamCudaPropagator object
+    MCSamples[iSample].Oscillator = new cudaprob3::BeamCudaPropagatorSingle(0, MCSamples[iSample].nEvents);
```
henry-israel, line 908: Currently I'm not sure if this will compile [it's giving error: expected type-specifier]

covariance/covarianceXsec.cpp
```diff
@@ -94,59 +112,321 @@ covarianceXsec::covarianceXsec(const char *name, const char *file, double thresh
...
+
+    if(xsec_param_names[i].length() > PrintLength) PrintLength = xsec_param_names.back().length();
     // DB Fill the stepscales vector
-    xsec_stepscale_vec[i] = (*xsec_stepscale)(i);    
+    xsec_stepscale_vec[i] = _fIndivStepScale[i];
 
+    if(_fFlatPrior[i]){setEvalLikelihood(i, true);}
+    
     for (int j = 0; j < ncols; j++) {
       xsec_param_id_a[i][j] = (*xsec_param_id)(i,j);
```
pgranger23, line 135: `xsec_param_id` is not initialized anymore, as L73 is commented. I am not sure how this should be resolved, but it is unusable in its current state.

covariance/covarianceXsec.cpp
```diff
@@ -425,144 +701,43 @@ void covarianceXsec::scanParameters() {
...
-    if ((GetXSecParamID(i, 1) & 1)) {//Near affecting parameters
-      if (GetXSecParamID(i, 0) == -1) {//Near affecting norm pars
+		// Or a normalisation parameter
+	  } //End FarSplinePars
+	  else if (strcmp(GetXsecParamType(i), "Norm") == 0) {//FarNormPars
 		XsecNorms4 tmp_xsec;
+		tmp_xsec.name=GetParameterName(i);
 
-        tmp_xsec.name=GetParameterName(i);
+		tmp_xsec.modes=_fNormModes[i];
```
pgranger23, line 738: Should rather be `tmp_xsec.modes=_fNormModes[norm_counter];` I think

covariance/covarianceXsec.h
EdAtkin: I haven't got round to validation the PCA with the YAML-ified systematics. This should be pretty easy but for now I have skipped it.

covariance/covarianceXsec.h
EdAtkin: yep, this just needs to be fixed in the spline generation tools to be consistent.

manager/manager.h
EdAtkin: Yeah the manager class still needs a thing and how we handle "core" things. I imagine this will be done by passing the yaml file down to the base class as well in the derived class constructor and then the base class grabbing the information it needs.

samplePDF/Structs.h
EdAtkin: I've added a default statement to throw if there's an unrecognised option. Longer term this could move into the OscClass or something I agree. I think for now it's not **so** bad though.

covariance/covarianceBase.h
EdAtkin: yeah we can remove these later. I've removed the const where I was getting an ignored qualifier warning.

CMakeLists.txt
```diff
@@ -93,29 +133,47 @@ set( CMAKE_EXPORT_COMPILE_COMMANDS ON )
 add_compile_options(-I${CUDA_SAMPLES}/common/inc)
 add_compile_options(-g -O3 -Wextra)
 
-#Add MultiThread flags
+#Compile and link options (Should be in experiment project but copied in case CORE is built alone) 
+add_compile_options(-g -Wextra)
+add_compile_options(-fopenmp)
+add_link_options(-fopenmp)
```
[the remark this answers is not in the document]
EdAtkin: correct, this is no fixed :)

CMakeLists.txt
```diff
@@ -45,12 +51,38 @@ CPMFindPackage(
...
+    CPMFindPackage(
+        NAME CUDAProb3
+        GITHUB_REPOSITORY "mach3-software/CUDAProb3"
+        GIT_TAG "feature_cleanup"
+	    OPTIONS "GPU_ON ON"
+    )
+
+  endif()
+
+endif()
```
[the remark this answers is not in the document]
EdAtkin: Yep I agree that this would be nicer. I think this isn't essential for now though. Can we flag this as an issue in the issue tracker?

covariance/covarianceXsec.h
EdAtkin: done

covariance/covarianceXsec.h
EdAtkin: I've removed a lot of the consts from the getters just by siting through the unqualified warnings. I haven't made them inline but I dont' think this is so important

samplePDF/samplePDFFDBase.cpp
EdAtkin: I don't see this error? So I think it's ok? We / I need to do a rigorous check of all the compile flags though so this can be a check there that all the different oscillation calculators cause problems

samplePDF/CMakeLists.txt
EdAtkin: Are you happy if we leave this for now?

covariance/covarianceBase.h
EdAtkin: I don't see anything that is likely to cause a problem. I think we can sort this out on when we're adding in all of Henry's adaptive stuff.

samplePDF/CMakeLists.txt
KSkwarczynski: It's ugly but doesn't break any funcionality so yes, let's leave it

covariance/covarianceXsec.cpp
EdAtkin: yep I've fixed this :)

samplePDF/samplePDFFDBase.cpp
EdAtkin: I removed this. It's set to NULL in samplePDFFDBase.h

KSkwarczynski (approved):

PR #42, opened 2024-02-13
End of unga bunga cmake?

KSkwarczynski:
This has been tested both with CPU and GPU and compiles with both.
This is our fancy cmake screen
image
image
There are several improvments so quickly

- Flag unification
- Define CUDA architecture once
- Use CUDAProb3 as header only not shared libs (to define flags!!!)
- Proper interface/cmake target

image

Thanks to changes we have control over flags in CUDAProb3
image

We have proper autogenerated setup
image

Thus we properly tranfer flags to for example T2K
image

Here are targets and Config.cmake
[MaCh3Config.cmake.txt](https://github.com/mach3-software/MaCh3/files/14269240/MaCh3Config.cmake.txt)
[MaCh3Targets.cmake.txt](https://github.com/mach3-software/MaCh3/files/14269241/MaCh3Targets.cmake.txt)

Is it perfect? Well no CUDAProb3 is a mess and it is impossible to compile them at the same time as linker goes crazy due to
`nvlink error   : Multiple definition of '_ZN9cudaprob37physics15mix_data_deviceE' in 'CMakeFiles/OscClass.dir/__/_deps/cudaprob3-src/atmoscudapropagator.cu.o', first defined in 'CMakeFiles/OscClass.dir/__/_deps/cudaprob3-src/beamcudapropagator.cu.o' (target: sm_35)`

Unless somone fixed CUDAProb3 we wwill still have fragile unga bunga cmake...


libanwars:
Sorry for making the unga bunga CMake !!

splines/gpuSplineUtils.cu
```diff
@@ -29,8 +29,8 @@
 #define __N_SPLINES__ NSPLINES_ND280
 // EM: for OA2024:
 #else
+#define __N_SPLINES__ 160
```
EdAtkin, line 32: There is obviously no way around this for now and only ND280 currently uses the gpuSplineUtils so this isn't a problem. But just a comment for myself to remember that we should think if there is some way around this?

EdAtkin (approved):
wow! There is a lot here! All of the changes look very sensible and in particular the CMake changes and the YAML helpers are really great. Thanks for doing so much work on this!

splines/gpuSplineUtils.cu
KSkwarczynski: Easiest way is in T2K cmake set number of splines as argument you pass to cpm. This argument is being set in cmake and core will compile with some number.

This way each experiment will be able to have different number of splines but as long as number is known at compilation level we are fine.

Long term we might ocnsider to removing it completeyl but this might impact performance so will need some testing.

PR #56, opened 2024-04-11
Minor fixes for cmake and compilation

pgranger23:
Enforces the downloading of the yaml-cpp package (https://github.com/cpm-cmake/CPM.cmake?tab=readme-ov-file#cpm_download_all). It could previously be found without being actually used.
Also adding a missing include in [splines/gpuSplineUtils.cu](https://github.com/mach3-software/MaCh3/compare/minor_fix?expand=1#diff-22d38a50e9089b009d3e907b5c38b5b4f1a9963fdb65833fb07bd228ddd87d18)


KSkwarczynski (approved):
Thank you for fixing it

PR #58, opened 2024-04-19
Optimization of fillArray code

pgranger23:
## Summary
Proposed optimization of the `fillArray` code. As the used data structure relies on Structure of Arrays (SoA), it is more efficient to apply all the operations, one after the other, to all the events at the same time rather than looping on the events to apply those operations.
Also avoids the need to make string comparisons on each event loop.

## Interface changes
- The definition of a `virtual double* ReturnKinematicParameter(std::string KinematicParamter, int iSample) = 0;` function is required for all the experiments in order to return a pointer to the relevant array to use as Kinematic Parameter. Note that this requires the KineMaticParameters to be of type `double`.
- The `fillArray_MP` and `fillArray` functions have been merged. I think it is better for future maintenance to have a single function to deal with and not 2 alternative versions.

## Performance improvements
The following performance improvements have been witnessed with these changes. The relevant points to compare here are the red and blue ones. These performances have been obtained on the DUNE FD sample, with the GCC compiler and the use of `--march=native` compiler flag.
profiling

## Things to verify
Here are the items that should be verified before the possible approval of this PR:
- Verification that it exactly reproduces the results from the previous code.
- Check that the use of the `#pragma omp` commands are never an issue, even when not compiling with multi-processing
- Check that this in fact leads to performance improvements for different samples

### DIFF
--- a/covariance/covarianceBase.cpp
+++ b/covariance/covarianceBase.cpp
@@ -444,7 +444,7 @@ void covarianceBase::init(std::vector<std::string> YAMLFile)
     }
   }
 
-  //setName(YAMLFile[0].c_str());
+  setName("yaml_systs");
   //size = covMatrix->GetNrows();
   //MakePosDef(covMatrix);
   //setCovMatrix(covMatrix);
--- a/samplePDF/samplePDFFDBase.cpp
+++ b/samplePDF/samplePDFFDBase.cpp
@@ -396,8 +396,8 @@ void samplePDFFDBase::fillArray() {
     }
   }
 
-  delete local_samplePDFFD_array;
-  delete local_samplePDFFD_array_w2;
+  delete[] local_samplePDFFD_array;
+  delete[] local_samplePDFFD_array_w2;
 
   }
   return;
--- a/samplePDF/samplePDFFDBase.cpp
+++ b/samplePDF/samplePDFFDBase.cpp
@@ -165,6 +165,27 @@ bool samplePDFFDBase::IsEventSelected(std::vector< std::string > ParameterStr, s
   return true;
 }
 
+//Similar to the above functions but applies to the whole sample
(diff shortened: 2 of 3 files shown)


KSkwarczynski:
About your comment
Check that the use of the #pragma omp commands are never an issue, even when not compiling with multi-processing

Usually what we do is wrap pragma around preprocessor
```
#ifdef MULTITHREAD
#pragma omp parallel for
#endif
```

All multithreading flags etc. are ste here
https://github.com/mach3-software/MaCh3/blob/82d1fec9ab8db6597053d78eb7b383dd29b30400/CMakeLists.txt#L196

covariance/covarianceBase.cpp
```diff
@@ -444,7 +444,7 @@ void covarianceBase::init(std::vector<std::string> YAMLFile)
     }
   }
 
-  //setName(YAMLFile[0].c_str());
+  setName("yaml_systs");
```
KSkwarczynski, line 447: Why was this changed?
This in principle can lead to situation where two covariances have the same name which is not good.
If you don't want name of covariance to be driven by the name of yaml you can change constructor to:
```
covarianceBase::covarianceBase(std::vector<std::string> YAMLFile, std::string name)
setName(name);
```

KSkwarczynski: Good point.

I think setting name via constructor would be safest.
You can set default value to reduce potentail compilation error. But will leave it to up to you.

dbarrow257:
Not that I expect a large change in RAM based on the changes I can see - but it would be good to check that the RAM usage has not significantly expanded in the new approach. In the short/mid term, the RAM usage is still somewhat important for pushing as many chains onto as few nodes as possible.

KSkwarczynski:
I think at this point this PR diverged enough that merging is impossible and we will have to cherry pick features.
I am closing PR becasue it is confusing

Tagging @EdAtkin just in case

PR #60, opened 2024-04-30
doxygen ci

tdealtry:
Automatically builds & pushes HTML to https://mach3-software.github.io/MaCh3/index.html when there are changes to develop

Note that this doesn't give version history for the documentation - it simply stores the latest doxygen build. This feature could be added in a future PR, if there is a need for it

Do let me know if anything looks odd
- one thing I noticed is the MaCh3 logo in the page header is **way** too big... I will fix that
- I'm not compiling/uploading the PDF, since I personally don't think this is required. Do shout if you disagree


KSkwarczynski:
Cheers, this looks lovely

I agree agree logo is too big...

Could you add link to github.io somewhere in readme? For example around here
https://github.com/mach3-software/MaCh3/blob/761cdc168663a6cfe659e4e3bab0d939bf715273/README.md?plain=1#L9

KSkwarczynski (approved):
Cheers

PR #61, opened 2024-06-19
Feature small covariance fixes

henry-wallace-phys:
Two small tweaks. 
 - I noticed oscillation parameters weren't being given the correct priors so I've just made the prior-setter fully explicit
 - delta-cp could, hypothetically be thrown outside of [-2 pi, 2pi] and not be set correctly.  I've used a bit of modular arithmetic to prevent this.

Prior Setting : 
Screenshot 2024-06-19 at 13 46 41

dcp stays within the correct bounds

[dcp_trace.pdf](https://github.com/user-attachments/files/15901084/dcp_trace.pdf)


KSkwarczynski (approved):
Cheers

PR #73, opened 2024-07-03
Drag Race

KSkwarczynski:
# Pull request description:
Add drag race which allow to see how much time rouglhy we need for each sample PDF object or covariance object

## Examples:
DragRace

image

After update
image


tdealtry:
Nice!
I wonder if there can be a (config setable) argument to this function to run N _laps_, and print the average lap time (maybe min/max too). Just because there can be variations in timing calculations like these

KSkwarczynski:
> Nice! I wonder if there can be a (config setable) argument to this function to run N _laps_, and print the average lap time (maybe min/max too). Just because there can be variations in timing calculations like these

Thanks for the suggestion just added this

PR #80, opened 2024-07-21
Adaptation Handler

KSkwarczynski:
# Pull request description:
Currently All AMCMC related stuff in heavily integrated within CovarianceBase. This PR attempts to make full fledget adaption class which interfaces with Covariance base. If all adaptation is done via this handler then selcting different adaptation via polimorphism will be super easy.

There are still Adaptive funciton in covaraince. This will need to be adressed later. Will try do this when I have time again.

## Changes or fixes:
- Add new AdaptiveMCMCHandler
- Update docuematnion
- Security no idea why but let's pretend we are fancy...


covariance/AdaptiveMCMCHandler.cpp
```diff
@@ -0,0 +1,231 @@
...
+AdaptiveMCMCHandler::~AdaptiveMCMCHandler() {
+// ********************************************
+  if(adaptive_covariance != nullptr) {
+    delete adaptive_covariance;
+  }
+}
+
+
+// ********************************************
+void AdaptiveMCMCHandler::InitFromConfig(const YAML::Node& adapt_manager, const std::string& matrix_name_str, const int Npars) {
```
henry-wallace-phys, line 27: Given Npars won't change does it not make sense to make this a class attribute rather than calling it in every method?

covariance/covarianceBase.cpp
```diff
@@ -1442,213 +1419,52 @@ void covarianceBase::initialiseAdaption(const YAML::Node& adapt_manager){
...
-  // https://projecteuclid.org/journals/bernoulli/volume-7/issue-2/An-adaptive-Metropolis-algorithm/bj/1080222083.full
+// ********************************************
   // Updates adaptive matrix
   // First we update the total means
 
   // Skip this if we're at a large number of steps
-  if(total_steps>adaption_struct.end_adaptive_update || total_steps<adaption_struct.start_adaptive_update) return;
-
-  int steps_post_burn = total_steps - adaption_struct.start_adaptive_update;
+  if(total_steps>AdaptiveHandler.end_adaptive_update || total_steps<AdaptiveHandler.start_adaptive_update) return;
```
henry-wallace-phys, line 1455: Conceptually do we not want to leave this logic inside the adaption handler?

PR #88, opened 2024-08-21
Small Tweak to code scope

henry-wallace-phys:
# Pull request description:
Moves Bin options getter to public scope of samplePDFFDBase [no real reason for it not to be and can be useful in some applications]


KSkwarczynski (approved):
[no real reason] 😂

PR #90, opened 2024-08-23
Fix debug build

ewanwm:
# Pull request description:
Couple of tiny fixes I came across while trying to debug something else

## Changes or fixes:
two variables misnames which was stopping code from compiling in debug mode

## Examples:
NA


KSkwarczynski (approved):
Thanks for this

PR #93, opened 2024-08-27
General Plotting Tools

ewanwm:
# Pull request description:
Adds a new plotting library which should hopefully be nice and general and can be used by experiments only by editing some config files.

Plotting library pretty much consists of just some high level manager classes that the user interacts with which can read in output files from different fitters and store the data based on config file that tells them how to read the data. 

The idea is that the user should be able to just include the `PlottingManager` class in their script and it will deal with all of the annoying stuff for them: 
- Read in all the input files from whatever file using the `inputManager` and keep the data in a nice tidy format which can be accessed with various getter functions 
- parse all the command line arguments, allowing all plotting scripts to use one unified and tidy command line interface
- easily apply styles defined in a style config file
- deal with translating names of parameters and samples between fitters so no more endless hard coded stuff to deal with slight changes in names from one year to another or between fitters

Definitely a lot of room for improvement and I've tried to highlight some obvious ways to improve with doxygen `@todo`s but it's a start 

## Changes or fixes:
- New plotting/plottingUtils directory with all the plotting library code
- Moved the PlotLLH and GetPostFitParameters to the plotting/ directory and change them to use the new plotting library (new GetPostFitParameters could definitely use some work BUT it runs and makes plots so thats work for a future PR)
- Nothing here should have any effect on any of the fitting code 

## Examples:

Have tested the code on some T2K 2024 ND analysis outputs using T2K config files (which I can commit to t2k refactor branch):
[PlottingConfig.yaml.txt](https://github.com/user-attachments/files/16763458/PlottingConfig.yaml.txt)
[universalTranslator.yaml.txt](https://github.com/user-attachments/files/16763459/universalTranslator.yaml.txt)
[StyleConfig.yaml.txt](https://github.com/user-attachments/files/16763460/StyleConfig.yaml.txt)

plots: 
[LLHScan_Penalty.pdf](https://github.com/user-attachments/files/16763399/LLHScan_Penalty.pdf)
[LLHScan_Sample.pdf](https://github.com/user-attachments/files/16763402/LLHScan_Sample.pdf)
[LLHScan_Total.pdf](https://github.com/user-attachments/files/16763403/LLHScan_Total.pdf)
[LLHScan_bySample.pdf](https://github.com/user-attachments/files/16763406/LLHScan_bySample.pdf)
[PostFitParamPlots.pdf](https://github.com/user-attachments/files/16763404/PostFitParamPlots.pdf)

And the all new ridgeline plots:
[RidgePlots.pdf](https://github.com/user-attachments/files/16763405/RidgePlots.pdf)


Doc/Doxyfile
```diff
@@ -733,7 +733,7 @@ WARN_LOGFILE           =
 # spaces.
 # Note: If this tag is empty the current directory is searched.
 
-INPUT                  = mainpage.md ../ ../manager ../splines ../samplePDF ../OscClass ../mcmc ../Diagnostics ../Diagnostics/Diagnostics_utils ../covariance Plots/
+INPUT                  = mainpage.md ../ ../manager ../splines ../samplePDF ../OscClass ../mcmc ../Diagnostics ../plotting ../plotting/plottingUtils ../Diagnostics/Diagnostics_utils ../covariance Plots/
```
KSkwarczynski, line 736: Can you update this as well
https://github.com/mach3-software/MaCh3/blob/develop/cmake/Templates/Doxyfile.in

I know it may look confusing.

Both doxyfiles are identical except
https://github.com/mach3-software/MaCh3/blob/9b319d291c2bfba67fb3b1abdc6ab25367cc9cb9/cmake/Templates/Doxyfile.in#L41
.in take Software version and should replace actual Foxyfile

plotting/plottingUtils/CMakeLists.txt
```diff
@@ -0,0 +1,21 @@
+
+## build the base level library targets
+foreach(lib
+    inputManager
+    styleManager
+    plottingUtils
+    )
+    add_library( ${lib} ${lib}.cpp )
+    target_link_libraries( ${lib} ${ROOT_LIBRARIES} MaCh3::Manager )
+
```
KSkwarczynski, line 10: Isn't better to MaCh3::All instead of MaCh3::Manager?

plotting/README.md
```diff
@@ -0,0 +1,45 @@
...
+
+### styleManager 
+
+This is responsible for the styling of the plots produces. It contains the functionality to apply user specified styles, colour palettes and fancy names for parameters and samples.
+
+## Config Files
+
+The goal with MaCh3s plotting library is to be as flexible as possible and to abstract away all of the most annoying parts about plot making like reading in the data from fitter output files and keeping track of parameter names across different fitters, while allowing the user as much freedom as possible when it comes plot style and formatting . In order to do this, MaCh3s plotting library makes heavy use of config files so that the behaviour of the code can be adapted easily to individual experiments without having to totally reinvent the wheel and rewrite large ammounts of boilerplate code. In line with the rest of MaCh3, the plotting library uses [YAML](https://yaml.org/spec/1.2.2/) to specify its config files. There are a few different config files which control different parts of the plotting library. Some example yaml configs found in this directory.
```
KSkwarczynski, line 23: Can you add reference to this here: https://github.com/mach3-software/MaCh3/tree/feature_Plotting?tab=readme-ov-file#plotting-and-diagnostic

plotting/plottingUtils/inputManager.cpp
```diff
@@ -0,0 +1,632 @@
+#include "inputManager.h"
+
+// EM: will move this somewhere more sensible
+#define __BAD_FLOAT__ -999.999
+
```
KSkwarczynski, line 5: Two underscoreso are reserved for the compiler's internal. It is safer to avoid double underscores

plotting/PlottingConfig.yaml
```diff
@@ -0,0 +1,79 @@
...
+            "XSEC_PARAM_3"
+        ]
+    ]
+
+    EXAMPLE_GROUP_2: ["COOL NAME 2", [-4.0, 4.0], 
+        [
+            "XSEC_PARAM_4",
+            "XSEC_PARAM_5",
+        ]
+    ]
```
KSkwarczynski, line 79: Can you remove leftovers here?
https://github.com/mach3-software/MaCh3/blob/6f94a99a46429a05713eba80ee6441dcbfc4f047/Diagnostics/Diag_Config.yaml#L128-L485

We also have to move (later) your T2K yaml configs. This can wait

KSkwarczynski (commented):
I love it 👍

plotting/plottingUtils/CMakeLists.txt
ewanwm: I wasn't sure. plotting stuff only uses things from manager (only yaml helper and logging) so linking manager seemed more "accurate". 
But also going forward, if in future we wanted to use any plotting utility in any of the other core code e.g. to make some kind of diagnostic/debug plot I think using MaCh3::All here wouldn't allow that right? (I am far from a CMake expert so not sure what's best)

KSkwarczynski: I see. So you think in the future plotting could become part of MaCh3::All, it is not right now thus got confused.

I can see your input manager to be used by other fucntions etc.

I was thinking that plotting utils might use parts of [Structs.h](https://github.com/mach3-software/MaCh3/blob/develop/samplePDF/Structs.h) or [StatisticalUtils.h](https://github.com/mach3-software/MaCh3/blob/develop/mcmc/StatisticalUtils.h) for example

Let's leave it unresolved for now then

KSkwarczynski (approved):
Thanks a lot

PR #94, opened 2024-08-29
Feature python plotting

ewanwm:
# Pull request description:
Extended the plotting library to include a python interface for the manager classes using [pybind11](https://github.com/pybind/pybind11). This means that we can now generate a python plotting module that can use all the c++ based plotting objects to read various input files based on existing configs and spit out the data in python. We can then use whatever lovely non-root plotting library to make beautiful and super flexible plots. 

## Changes or fixes:
- Added high level pyMaCh3 module (see pyMaCh3.cpp) which is just kind of a dummy module (.... for now.) which holds plotting sub-module
- Added plotting sub-module (see plotting/plottingUtils/pythonPlottingModule.cpp) which binds all of the useful plotting library c++ manager objects to python

## Examples:
Used T2K OA2024 ND Asimov LLH scans and suuuuper simple python script:

```
from pyMaCh3.plotting import PlottingManager
import matplotlib.backends.backend_pdf
import sys
from matplotlib import pyplot as plt
import numpy as np


man = PlottingManager()

man.parse_inputs(sys.argv)

pdf = matplotlib.backends.backend_pdf.PdfPages("LLH-scans.pdf")

for i, param in enumerate( man.input().get_known_parameters() ):
    
    fig = plt.figure()
    plt.title(man.style().prettify_parameter_name(param))

    for file_id in range(man.input().get_n_input_files()):
        llh_scan = man.input().get_llh_scan(file_id, param, "sample")
        plt.plot(llh_scan[0], llh_scan[1])
    
    pdf.savefig( fig )

pdf.close()
```

to generate some plots with matplotlib:
[LLH-scans.pdf](https://github.com/user-attachments/files/16801567/LLH-scans.pdf)


KSkwarczynski (approved):
Spicy :)

it would be good if @henry-wallace-phys could also comment as pythony expert

ewanwm:
(I did also speak to henry separately. I realise now it looks like i just ignored that last comment)

PR #154, opened 2024-10-07
save the 1d posteriors in MCMCProcessor

ewanwm:
# Pull request description:

Does what it says on the tin really


KSkwarczynski (approved):
Good idea, thanks

PR #157, opened 2024-10-09
Feature fd constructor cleanup

EdAtkin:
# Pull request description:
Changes to move a lot of the samplePDFFDBase constructor now happen in core. 

## Changes or fixes:
Previously lots of this had to be done in the experiment specific code but this is now in core with the help of a few virtual functions to sort out the string to kinematic parameter mapping and vice versa.

A few extra little bits here and there such as logger messages and some doxygen comments have been added occasionally.

## Examples:
ReadSampleConfig in samplePDFFDBase now does all of the reading from the sample config which is necessary.


KSkwarczynski:
Closing this becasue of 
https://github.com/mach3-software/MaCh3/pull/161

PR #161, opened 2024-10-10
Feature osc class magic

EdAtkin:
# Pull request description
Contains both samplePDFFD clean up changes and uses the NuOscillator class for oscillation calculation.

## Changes or fixes
Moves a lot of the samplePDFFD constructor into core which simplifies setting up a derived class.
Removed of a lot of the boiler plate for oscillation weights and calculations since this is now done in NuOscillator.


samplePDF/samplePDFFDBase.h
```diff
@@ -238,18 +231,40 @@ class samplePDFFDBase :  public samplePDFBase
...
+  std::vector< std::string > SelectionStr; 
+  /// @brief the bounds for each selection lower and upper
+  std::vector< std::vector<double> > SelectionBounds;
+  /// @brief a way to store selection cuts which you may push back in the get1DVar functions
+  /// most of the time this is just the same as StoredSelection
+  std::vector< std::vector<double> > Selection;
+   //===========================================================================
+
+  manager* SampleManager;
+  void InitialiseSingleFDMCObject(int iSample, int nEvents);
```
KSkwarczynski, line 252: I don't have strong opinion but do we need manager* or YAML::NODE woulbe be enough?

samplePDF/samplePDFFDBase.cpp
```diff
@@ -1497,27 +1532,96 @@ double samplePDFFDBase::GetLikelihood()
...
-    case CUDAProb_nu::t_m : return cudaprob3::ProbType::t_m;
-    case CUDAProb_nu::t_t : return cudaprob3::ProbType::t_t;
-    default:
-      MACH3LOG_ERROR("Unknown CUDAProbType!");
-      throw;
+
+void samplePDFFDBase::InitialiseSingleFDMCObject(int iSample, int nEvents_) {
+  if (iSample < 0 || iSample >= nSamples) {
+    MACH3LOG_ERROR("Invalid iSample index in InitialiseSingleFDMCObject");
+    MACH3LOG_ERROR("Index given is {} and only {} samples found", iSample, nSamples);
```
KSkwarczynski, line 1539: can we also throw this away:

https://github.com/mach3-software/MaCh3/blob/feature_OscClassMagic/samplePDF/Structs.h#L513-L581

samplePDF/samplePDFFDBase.cpp
```diff
@@ -1414,11 +1358,103 @@ void samplePDFFDBase::addData(TH2D* Data) {
   }
 }
 
+void samplePDFFDBase::SetupNuOscillator() {
+  OscProbCalcerFactory* OscProbCalcFactory = new OscProbCalcerFactory();
```
KSkwarczynski, line 1362: OscProbCalcFactory  is not used fron what I see

samplePDF/samplePDFFDBase.cpp
```diff
@@ -1414,11 +1358,103 @@ void samplePDFFDBase::addData(TH2D* Data) {
...
+  OscillatorFactory* OscillFactory = new OscillatorFactory();
```
KSkwarczynski, line 1363: OscillFactory is not deleted :(

samplePDF/samplePDFFDBase.cpp
```diff
@@ -1414,11 +1358,103 @@ void samplePDFFDBase::addData(TH2D* Data) {
...
+  
+  NuOscProbCalcers = std::vector<OscillatorBase*>((int)MCSamples.size());
+  for (int iSample=0;iSample<(int)MCSamples.size();iSample++) {
```
KSkwarczynski, line 1366: NuOscProbCalcers is not deleted in destructor :(

samplePDF/samplePDFFDBase.h
```diff
@@ -238,18 +231,40 @@ class samplePDFFDBase :  public samplePDFBase
...
+  std::vector<int> sample_vecno;
+  std::vector<int> sample_oscnutype;
+  std::vector<int> sample_nutype;
+  std::vector<bool> sample_signal;
+
+  std::string mtupleprefix;
+  std::string mtuplesuffix;
+  std::string splineprefix;
+  std::string splinesuffix;
+
```
KSkwarczynski, line 269: Are these needed?
https://github.com/mach3-software/MaCh3/blob/5fa87ded750b3ba2f040e7eddd9e51e7dcc7350b/samplePDF/samplePDFFDBase.cpp#L135
Here you use them locally.

So stuff defined in header is really useless

samplePDF/samplePDFFDBase.cpp
```diff
@@ -1,37 +1,34 @@
...
+  //ETA - safety feature so you can't pass a NULL xsec_cov
+  if(xsec_cov == NULL){
+	MACH3LOG_ERROR("You've passed me a NULL xsec covariance matrix... I need this to setup splines!");
+   	throw MaCh3Exception(__FILE__, __LINE__);
+  }
+  SetXsecCov(xsec_cov);
+  
+  samplePDFFD_array = NULL;
+  samplePDFFD_data = NULL;
+
```
KSkwarczynski, line 29: can we use nullptr, it is genrally recomended for modern c++
https://stackoverflow.com/a/20509811

samplePDF/samplePDFFDBase.cpp
```diff
@@ -1,37 +1,34 @@
...
+#include "OscProbCalcer/OscProbCalcerFactory.h"
+#include "Oscillator/OscillatorFactory.h"
+#include "Constants/OscillatorConstants.h"
 
-samplePDFFDBase::samplePDFFDBase(double pot, std::string mc_version, covarianceXsec* xsec_cov)
+#include<algorithm>
+
+
+samplePDFFDBase::samplePDFFDBase(double pot, std::string mc_version_, covarianceXsec* xsec_cov)
   : samplePDFBase()
```
KSkwarczynski, line 11: Do we need to pass pot ?
Doesn't it make more sense to store POT in sample config?

KSkwarczynski (commented):
NuOscillator is coming home!

samplePDF/samplePDFFDBase.cpp
```diff
@@ -1414,11 +1358,103 @@ void samplePDFFDBase::addData(TH2D* Data) {
...
+	for (int iEvent=0;iEvent<(int)MCSamples[iSample].nEvents;iEvent++) {
+	  CosineZArray[iEvent] = *(MCSamples[iSample].rw_truecz[iEvent]);
+	}
+	std::sort(CosineZArray.begin(),CosineZArray.end());
+	
+	NuOscProbCalcers[iSample]->SetCosineZArrayInCalcer(CosineZArray);
+      }
+      //============================================================================
+      
+    }
```
KSkwarczynski, line 1393: As complete newbie should't here be check for NC events?
Otherwise you have larger vector and I htink NuOsc will calclatue for NC events. 
Weights will be correctly assgiend it just about SPEEEED

samplePDF/samplePDFFDBase.h
dbarrow257: manager would be better because we use CheckNodeExists functions: https://github.com/mach3-software/MaCh3/blob/5fa87ded750b3ba2f040e7eddd9e51e7dcc7350b/samplePDF/samplePDFFDBase.cpp#L56

samplePDF/samplePDFFDBase.cpp
dbarrow257: No that is correct, this can be removed (It was from previous implementation where I was just directly dealing with OscProbCalcerBase objects)

samplePDF/samplePDFFDBase.cpp
dbarrow257: It's a hangover from previous stuff that should definitely be cleaned...

It even get's void-ed in the base constructor: https://github.com/mach3-software/MaCh3/blob/5fa87ded750b3ba2f040e7eddd9e51e7dcc7350b/samplePDF/samplePDFFDBase.cpp#L10-L14

samplePDF/samplePDFFDBase.cpp
dbarrow257: Yeah sure, I guess we could add check for NC events and remove from array - we're just lazy...

NC events don't get assigned any probabilities: https://github.com/mach3-software/MaCh3/blob/5fa87ded750b3ba2f040e7eddd9e51e7dcc7350b/samplePDF/samplePDFFDBase.cpp#L1399-L1404

So yes, we should make an effort to remove NC event energies/cosines from the array

samplePDF/samplePDFFDBase.h
KSkwarczynski: Check node exist isn't part of manager but YamlHelper.h
https://github.com/mach3-software/MaCh3/blob/60dfdff0c2e1c709161779eb5d1f23b2e280a3b4/manager/YamlHelper.h#L56

So you can just pass yaml node.

Again I don't have strong opinions about this one

dbarrow257: Ah okay, seems like there's no strong reason to move it to a YAML::Node

samplePDF/samplePDFFDBase.h
dbarrow257: The spline prefixes are actually used here; https://github.com/mach3-software/MaCh3/blob/5fa87ded750b3ba2f040e7eddd9e51e7dcc7350b/samplePDF/samplePDFFDBase.cpp#L1606

samplePDF/samplePDFFDBase.cpp
EdAtkin: this is now done https://github.com/mach3-software/MaCh3/blob/feature_OscClassMagic/samplePDF/samplePDFFDBase.cpp#L46

samplePDF/samplePDFFDBase.cpp
dbarrow257: This has now been removed and should be added to any beam-only samples as a member variable

samplePDF/samplePDFFDBase.cpp
dbarrow257: NC events now removed from the array which is passed to the NuOscillator objects

samplePDF/samplePDFFDBase.cpp
```diff
@@ -51,6 +41,168 @@ samplePDFFDBase::~samplePDFFDBase()
...
+  return;
+}
+
+void samplePDFFDBase::Initialise() {
+
+  //First grab all the information from your sample config via your manager
+  ReadSampleConfig();
+
+  //Now initialise all the variables you will need
+  Init();
```
luketpickering, line 177: a function named `Init` in a function named `Initialise`, that looks like it should be part of the constructor, is confusing. It appears that this is the virtual call to tell subclasses to allocate memory that we are about to use. This tightly ordered and dependent initialisation sequence looks like it is neccessary, but it can be simplified.

Subclasses should perform their `Init` in their constructors, and then this `Initialise` function should be named `InitialiseBase` or something clear.

If the ordering could be uncoupled so that the subclass can initialise last, but still all be done in the subclass constructor, that would be nice, but I suspect it can't, so the flow here is almost the clearest it can be, but having functions named `Init` or similar that are just providing RAII/construction logic is confusing.

If you need multiple constructor signatures that differ but have common procedures, you can use delegating constructors: https://www.ibm.com/docs/en/zos/2.4.0?topic=only-delegating-constructors-c11

samplePDF/samplePDFFDBase.cpp
```diff
@@ -100,105 +253,77 @@ void samplePDFFDBase::SetupSampleBinning(){
...
-  XBinEdges.reserve(SampleNXBins);
-  YBinEdges.reserve(SampleNYBins);
-  std::cout << "XBinning: " << std::endl;
-  for(unsigned XBin_i = 0 ; XBin_i < SampleNXBins ; XBin_i++){
-	XBinEdges.push_back(SampleXBins[XBin_i]);
-	std::cout << SampleXBins[XBin_i] << ", ";
-  }
-  std::cout << "\n" << std::endl;
+  XBinEdges.reserve(SampleXBins.size());
+  YBinEdges.reserve(SampleYBins.size());
```
luketpickering, line 257: unless you're in a very tight loop, reserve is usually just noise.

samplePDF/samplePDFFDBase.cpp
```diff
@@ -100,105 +253,77 @@ void samplePDFFDBase::SetupSampleBinning(){
...
-  //And now the YBin Edges
-  std::cout << "YBinning: " << std::endl;
-  for(unsigned YBin_i = 0 ; YBin_i < SampleNYBins ; YBin_i++){
-	YBinEdges.push_back(SampleYBins[YBin_i]);
-	std::cout << SampleYBins[YBin_i] << ", ";
+  //A string to store the binning for a nice print out
+  std::string XBinEdgesStr = "";
+  std::string YBinEdgesStr = "";
+
+  for(auto XBinEdge : SampleXBins){
```
luketpickering, line 263: I think this is unnecessary. spdlog doesn't seem to include all of the fmtlib features, but we can use its bundled fmtlib.

```c++
//use spdlog bundled fmtlib ranges
#include "spdlog/fmt/bundled/ranges.h"

//...

MACH3LOG_INFO(fmt::format("XBinning: {}",SampleXBins));
```

samplePDF/samplePDFFDBase.cpp
```diff
@@ -100,105 +253,77 @@ void samplePDFFDBase::SetupSampleBinning(){
...
-    //DB In the case where Selection[0].size()==3, Only Events with Val >= Selection[iSelection][1] and Val < Selection[iSelection][2] are considered Passed
     if ((Val<Selection[iSelection][1])||(Val>=Selection[iSelection][2])) {
       return false;
     }
-
   }
-
-  //DB To avoid unneccessary checks, now return false rather than setting bool to true and continuing to check
+  
+  //DB To avoid unnecessary checks, now return false rather than setting bool to true and continuing to check
```
luketpickering, line 310: the comment says that it will return false, but it doesn't, is it in the wrong place or outdated?

samplePDF/samplePDFFDBase.cpp
```diff
@@ -100,105 +253,77 @@ void samplePDFFDBase::SetupSampleBinning(){
...
+    Val = ReturnKinematicParameter(ParameterStr[iSelection], iSample, iEvent);
     if ((Val<SelectionBounds[iSelection][0])||(Val>=SelectionBounds[iSelection][1])) {
-	  return false;
+      return false;
     }
   }
-
-  //DB To avoid unneccessary checks, now return false rather than setting bool to true and continuing to check
+  
+  //DB To avoid unnecessary checks, now return false rather than setting bool to true and continuing to check
```
luketpickering, line 326: and here

samplePDF/samplePDFFDBase.cpp
```diff
@@ -207,120 +332,55 @@ bool samplePDFFDBase::IsEventSelected(const std::vector< std::string >& Paramete
...
-    //DB In the case where SelectionCuts[0].size()==3, Only Events with Val >= SelectionCuts[iSelection][1] and Val < SelectionCuts[iSelection][2] are considered Passed
-	//ETA - also check whether we're actually applying a lower or upper cut by checking they aren't -999
-	if(Val >= SelectionCuts[iSelection][1] && SelectionCuts[iSelection][0] != -999){
-	  //std::cout << "Cutting event as " << Val << " is greater than " << SelectionCuts[iSelection][1]
-	  return false;
-	}
-	else if(Val < SelectionCuts[iSelection][0] && SelectionCuts[iSelection][1] != -999){
-	  return false;
-	}
+    if(Val >= SelectionCuts[iSelection][1] && SelectionCuts[iSelection][0] != -999){
```
luketpickering, line 341: why -999?

I think there are macros (yuck) of 'BAD_FLOAT' etc, defined, so I assume its that, but then you should check equality with the macro value not a hard coded value that happens to be the same as the macro at some point.

-999 is also not a great bad value, not the worst, but not great.

samplePDF/samplePDFFDBase.cpp
```diff
@@ -207,120 +332,55 @@ bool samplePDFFDBase::IsEventSelected(const std::vector< std::string >& Paramete
...
+      return false;
+    }
+    else if(Val < SelectionCuts[iSelection][0] && SelectionCuts[iSelection][1] != -999){
+      return false;
+    }
   }
-
-  //DB To avoid unneccessary checks, now return false rather than setting bool to true and continuing to check
+  
+  //DB To avoid unnecessary checks, now return false rather than setting bool to true and continuing to check
```
luketpickering, line 349: and here

samplePDF/samplePDFFDBase.cpp
```diff
@@ -207,120 +332,55 @@ bool samplePDFFDBase::IsEventSelected(const std::vector< std::string >& Paramete
...
-void samplePDFFDBase::calcOscWeights(int sample, int nutype, double *w)
-{
-  MCSamples[sample].Oscillator->setMNSMatrix(asin(sqrt(*oscpars[0])),asin(sqrt(*oscpars[2])), asin(sqrt(*oscpars[1])), (*oscpars[5]), nutype);
-  MCSamples[sample].Oscillator->setNeutrinoMasses(*oscpars[3], *oscpars[4]);
-  MCSamples[sample].Oscillator->calculateProbabilities(MCSamples[sample].NeutrinoType);
-  MCSamples[sample].Oscillator->getProbabilityArr(w, MCSamples[sample].ProbType);
-}
-#endif 
-
 void samplePDFFDBase::reweight() // Reweight function - Depending on Osc Calculator this function uses different CalcOsc functions
```
luketpickering, line 353: this comment is unclear. Maybe:

```
Applies oscillation weights (and systematic weights???), uses the currently enabled Osc Calculator to calculate oscillation weights.
```

samplePDF/samplePDFFDBase.cpp
```diff
@@ -207,120 +332,55 @@ bool samplePDFFDBase::IsEventSelected(const std::vector< std::string >& Paramete
...
-      calcOscWeights(MCSamples[i].nutype, MCSamples[i].oscnutype, *(MCSamples[i].rw_etru), MCSamples[i].osc_w, MCSamples[i].nEvents);
-#endif
-      
-#if not defined (USE_PROB3)
-      calcOscWeights(i, MCSamples[i].nutype, MCSamples[i].osc_w);
-#endif
-    }// Sample loop
-  }
-
   //KS: Reset the histograms before reweight
```
luketpickering, line 355: thats what the function name says, we  shouldn't add noisey comments that write out function names longform, we should comment why that operation needs to be done. Why do we need to reset them? what does resetting them mean/do? set the bin contents to 0? nominal? do I need to know to understand the following code?

samplePDF/samplePDFFDBase.cpp
```diff
@@ -207,120 +332,55 @@ bool samplePDFFDBase::IsEventSelected(const std::vector< std::string >& Paramete
...
 }
 
-// ************************************************
-//DB Function which does the core reweighting. This assumes that oscillation weights have already been calculated and stored in samplePDFFDBase[iSample].osc_w[iEvent]
-//This function takes advantage of most of the things called in setupSKMC to reduce reweighting time
-//It also follows the ND code reweighting pretty closely
-//This function fills the samplePDFFD_array array which is binned to match the sample binning, such that bin[1][1] is the equivalent of _hPDF2D->GetBinContent(2,2) {Noticing the offset}
+//************************************************
+/// @function samplePDFFDBase::fillArray()
+/// Function which does the core reweighting. This assumes that oscillation weights have
```
luketpickering, line 373: why is the function that does the core reweighting called `fillArray` and there is another function called `reweight`, which apparently doesn't do the core reweighting? or does some part of it?

samplePDF/samplePDFFDBase.cpp
```diff
@@ -340,328 +400,289 @@ void samplePDFFDBase::fillArray() {
...
       //DB Catch negative weights and skip any event with a negative event
       if (totalweight <= 0.){
-		MCSamples[iSample].xsec_w[iEvent] = 0.;
-	   	continue;
-	  }
- 
+        MCSamples[iSample].xsec_w[iEvent] = 0.;
+        continue;
+      }
       //DB Switch on BinningOpt to allow different binning options to be implemented
```
luketpickering, line 453: is this comment still relevant?

samplePDF/samplePDFFDBase.cpp
```diff
@@ -340,328 +400,289 @@ void samplePDFFDBase::fillArray() {
...
       //The alternative would be to have inheritance based on BinningOpt
       double XVar = *(MCSamples[iSample].x_var[iEvent]);
-
+      
       //DB Find the relevant bin in the PDF for each event
       int XBinToFill = -1;
```
luketpickering, line 458: seems that shift systematics do only work in X, is that clearly documented somewhere?

samplePDF/samplePDFFDBase.cpp
```diff
@@ -340,328 +400,289 @@ void samplePDFFDBase::fillArray() {
...
   // Declare the omp parallel region
   // The parallel region needs to stretch beyond the for loop!
 #pragma omp parallel private(samplePDFFD_array_private, samplePDFFD_array_private_w2)
   {
-	// private to each thread
-	// ETA - maybe we can use parallel firstprivate to initialise these?
-	samplePDFFD_array_private = new double*[nYBins];
+    // private to each thread
+    // ETA - maybe we can use parallel firstprivate to initialise these?
+    samplePDFFD_array_private = new double*[nYBins];
```
luketpickering, line 528: use vectors pls

samplePDF/samplePDFFDBase.cpp
```diff
@@ -340,328 +400,289 @@ void samplePDFFDBase::fillArray() {
...
+	if (totalweight <= 0.){
+	  MCSamples[iSample].xsec_w[iEvent] = 0.;
+	  continue;
+	}
+	
+	//DB Switch on BinningOpt to allow different binning options to be implemented
+	//The alternative would be to have inheritance based on BinningOpt
+	double XVar = (*(MCSamples[iSample].x_var[iEvent]));
+	
+	//DB Commented out by default but if we ever want to consider shifts in theta this will be needed
```
luketpickering, line 622: eeek re: `theta`

samplePDF/samplePDFFDBase.cpp
```diff
@@ -340,328 +400,289 @@ void samplePDFFDBase::fillArray() {
...
+  
 #endif // end the else in openMP
   return;
 }
 
 #ifdef MULTITHREAD
+// ************************************************ 
+/// Multithreaded version of fillArray @see fillArray()
+// ************************************************ 
 void samplePDFFDBase::fillArray_MP()
```
luketpickering, line 504: this is so similar to the 'single threaded' version, I think it would be worth trying to reduce code duplication

samplePDF/samplePDFFDBase.cpp
```diff
@@ -340,328 +400,289 @@ void samplePDFFDBase::fillArray() {
...
+    }    
+    
+    //End of Calc Weights and fill Array
+    //==================================================
+    // DB Copy contents of 'samplePDFFD_array_private' into 'samplePDFFD_array' which can then be used in GetLikelihood
+    for (int yBin=0;yBin<nYBins;yBin++) {
+      for (int xBin=0;xBin<nXBins;xBin++) {
 #pragma omp atomic
-		  samplePDFFD_array[yBin][xBin] += samplePDFFD_array_private[yBin][xBin];
+	samplePDFFD_array[yBin][xBin] += samplePDFFD_array_private[yBin][xBin];
```
luketpickering, line 675: its probably cleaner to have each thread fill its own array that allocated before the parallel region and then after the join, let a single thread add them all up. This isn't a speedy-needing region, but atomics introduce introduce overheads, so its not immediately clear (though maybe you've profiled), that this will be faster when N threads are all fighting to add to the same double, rather than just letting them all write to their own memory and having a single thread do the final sum.

samplePDF/samplePDFFDBase.cpp
```diff
@@ -806,171 +800,110 @@ void samplePDFFDBase::CalcXsecNormsBins(int iSample){
...
+	      ModeMatch=true;
+	    }
+	  }
+	}
+	if (!ModeMatch) {continue;}
+	
+	//Now check whether the norm has kinematic bounds
+	//i.e. does it only apply to events in a particular kinematic region?
+	bool IsSelected = true;
+	if ((*it).hasKinBounds) {
```
luketpickering, line 872: its so minor, but I really don't understand the `(*it).`, why not just use the syntax that everyone understands/learns in C++ 101, `it->hasKinBounds`?

samplePDF/samplePDFFDBase.cpp
```diff
@@ -991,14 +924,14 @@ void samplePDFFDBase::set1DBinning(std::vector<double> &XVec){
   samplePDFFD_array = new double*[nYBins];
```
luketpickering, line 924: vectors

samplePDF/samplePDFFDBase.cpp
```diff
@@ -1239,94 +1171,79 @@ void samplePDFFDBase::FindNominalBinAndEdges2D() {
...
-	  MCSamples[mc_i].rw_lower_xbinedge[event_i] = low_edge;
-	  MCSamples[mc_i].rw_upper_xbinedge[event_i] = upper_edge;
-	  MCSamples[mc_i].rw_lower_lower_xbinedge[event_i] = low_lower_edge;
-	  MCSamples[mc_i].rw_upper_upper_xbinedge[event_i] = upper_upper_edge;
-	}
+      
+      //Global bin number
+      int bin = _hPDF2D->FindBin(*(MCSamples[mc_i].x_var[event_i]), *(MCSamples[mc_i].y_var[event_i]));
+      
+      int bin_x = -999;
```
luketpickering, line 1183: why -999 again? use a constant if it is a magic value

samplePDF/samplePDFFDBase.cpp
```diff
@@ -1239,94 +1171,79 @@ void samplePDFFDBase::FindNominalBinAndEdges2D() {
...
-	std::cerr << "Error, unrecognsied BinningOpt!!" << std::endl;
-	throw;
-    return 0;
-  }  
-}
-*/
-
 void samplePDFFDBase::addData(std::vector<double> &data) {
   dataSample = new std::vector<double>(data);
   dataSample2D = NULL;
```
luketpickering, line 1230: `nullptr` if you have to use pointers

samplePDF/samplePDFFDBase.cpp
```diff
@@ -1239,94 +1171,79 @@ void samplePDFFDBase::FindNominalBinAndEdges2D() {
...
-  default:
-	std::cerr << "Error, unrecognsied BinningOpt!!" << std::endl;
-	throw;
-    return 0;
-  }  
-}
-*/
-
 void samplePDFFDBase::addData(std::vector<double> &data) {
   dataSample = new std::vector<double>(data);
```
luketpickering, line 1229: `std::vector<std::vector<double>>` is fine

samplePDF/samplePDFFDBase.cpp
luketpickering: but like...

```c+++
auto OscillFactory = std::make_unique<OscillatorFactory>();
```

is less effort, less characters, more sparkles

samplePDF/samplePDFFDBase.cpp
```diff
@@ -1414,11 +1333,118 @@ void samplePDFFDBase::addData(TH2D* Data) {
...
+      //============================================================================
+      
+    }
+    
+    NuOscProbCalcers[iSample]->Setup();
+
+    for (int iEvent=0;iEvent<(int)MCSamples[iSample].nEvents;iEvent++) {
+      // KS: Sry but if we use low memory we need to point to float not double...
+#ifdef _LOW_MEMORY_STRUCTS_
+      MCSamples[iSample].osc_w_pointer[iEvent] = &Unity_F;
```
luketpickering, line 1379: its better to just change how `Unity` is declared based on this define, rather than littering the code with hard to parse CPP directives

samplePDF/samplePDFFDBase.cpp
```diff
@@ -1428,96 +1454,166 @@ double samplePDFFDBase::GetEventWeight(int iSample, int iEntry) {
...
+  fdmc_base *fdobj = &MCSamples[iSample];
+  
+  fdobj->nEvents = nEvents_;
+  fdobj->nutype = -9;
+  fdobj->oscnutype = -9;
+  fdobj->signal = false;
+  fdobj->Unity = 1.;
+  fdobj->Unity_Int = 1.;
+  
+  fdobj->x_var = new const double*[fdobj->nEvents];
```
luketpickering, line 1540: vectors

samplePDF/samplePDFFDBase.cpp
```diff
@@ -1428,96 +1454,166 @@ double samplePDFFDBase::GetEventWeight(int iSample, int iEntry) {
...
+  fdobj->rw_lower_xbinedge = new double [fdobj->nEvents];
+  fdobj->rw_lower_lower_xbinedge = new double [fdobj->nEvents];
+  fdobj->rw_upper_xbinedge = new double [fdobj->nEvents];
+  fdobj->rw_upper_upper_xbinedge = new double [fdobj->nEvents];
+  fdobj->mode = new int*[fdobj->nEvents];
+  fdobj->nxsec_norm_pointers = new int[fdobj->nEvents];
+  fdobj->xsec_norm_pointers = new const double**[fdobj->nEvents];
+  fdobj->xsec_norms_bins = new std::list< int >[fdobj->nEvents];
+  fdobj->xsec_w = new double[fdobj->nEvents];
+  fdobj->isNC = new bool[fdobj->nEvents];
```
luketpickering, line 1558: DONT USE `std::vector<bool>` they broke it and it won't do what you expect

samplePDF/samplePDFFDBase.cpp
```diff
@@ -1428,96 +1454,166 @@ double samplePDFFDBase::GetEventWeight(int iSample, int iEntry) {
...
+
+    fdobj->x_var[iEvent] = &fdobj->Unity;
+    fdobj->y_var[iEvent] = &fdobj->Unity;
+  }
+
 }
-#endif
 
+void samplePDFFDBase::InitialiseSplineObject() {
+  std::vector<std::string> spline_filepaths;
```
luketpickering, line 1596: can this code not be moved to the constructor of the spline object? or would there be too many arguments to pass? Can the arguments be wrapped up in a 'config struct' that can be passed to spline objects to initialise themselves?

samplePDF/samplePDFFDBase.cpp
```diff
@@ -1428,96 +1454,166 @@ double samplePDFFDBase::GetEventWeight(int iSample, int iEntry) {
...
+  //Keep a track of the spline variables
+  SplineVarNames.push_back("TrueNeutrinoEnergy");
+  if(XVarStr.length() > 0){
+    SplineVarNames.push_back(XVarStr);
+  }
+  if(YVarStr.length() > 0){
+    SplineVarNames.push_back(YVarStr);
+  }
+  
+  splineFile->AddSample(samplename, SampleDetID, spline_filepaths, SplineVarNames);
```
luketpickering, line 1610: minor: this object is not a file, so the variable handle shouldn't be `splineFile`

samplePDF/samplePDFFDBase.h
```diff
@@ -17,7 +18,6 @@
 
 #include "samplePDF/samplePDFBase.h"
 #include "samplePDF/FDMCStruct.h"
-#include "samplePDF/ShiftFunctors.h"
 
 /// @brief Class responsible for handling implementation of samples used in analysis, reweighting and returning LLH
 class samplePDFFDBase :  public samplePDFBase
```
luketpickering, line 23: does it need to have FD in the name? does anything subclass `samplePDFBase` directly? Can they be merged?

samplePDF/samplePDFFDBase.h
```diff
@@ -52,27 +50,14 @@ class samplePDFFDBase :  public samplePDFBase
 
   // Setup and config functions
   void UseNonDoubledAngles(bool ans) {doubled_angle = ans;};
```
luketpickering, line 52: isn't this a property of the oscillation calculator, rather than the samplePDF?

samplePDF/samplePDFFDBase.h
```diff
@@ -52,27 +50,14 @@ class samplePDFFDBase :  public samplePDFBase
...
-  inline std::string GetName() const override { return samplename; }
-
   const double **oscpars;
   void SetXsecCov(covarianceXsec* xsec_cov);
   void SetOscCov(covarianceOsc* osc_cov);
 
+  ///  @brief including Dan's magic NuOscillator
+  void SetupNuOscillator();
+
   /// @deprecated The `DumpWeights` function is deprecated and should not be used.
```
luketpickering, line 61: pls delete it if its not needed. move fast and break things

samplePDF/samplePDFFDBase.h
```diff
@@ -84,26 +69,32 @@ class samplePDFFDBase :  public samplePDFBase
...
+  //DB Experiment specific setup, returns the number of events which were loaded
+  virtual int setupExperimentMC(int iSample) = 0;
+
+  //DB Function which translates experiment struct into core struct
+  virtual void setupFDMC(int iSample) = 0;
+
+  //DB Function which does a lot of the lifting regarding the workflow in creating different MC objects
+  void Initialise();
   
   splineFDBase *splineFile;
```
luketpickering, line 96: not a file, name could be better

samplePDF/samplePDFFDBase.h
```diff
@@ -134,8 +125,9 @@ class samplePDFFDBase :  public samplePDFBase
   bool IsEventSelected(const std::vector<std::string>& ParameterStr, const int iSample, const int iEvent);
   bool IsEventSelected(const std::vector<std::string>& ParameterStr, const std::vector<std::vector<double>> &SelectionCuts, const int iSample, const int iEvent);
 
+  /// @brief Check whether a normalisation systematic affects an event or not
   void CalcXsecNormsBins(int iSample);
-  /// @brief This just gets read in from a yaml file
+  /// @brief Is the sample for when operating in Reverse Horn Current, read in from sample config
```
luketpickering, line 130: pet peeve: what does RHC mean? better to refer to neutrino-mode and antineutrino-mode. Doesn't require people to remember which charge is focussed depending on the current flow through the horn conductors.

samplePDF/samplePDFFDBase.h
```diff
@@ -147,10 +139,12 @@ class samplePDFFDBase :  public samplePDFBase
   virtual double ReturnKinematicParameter(std::string KinematicParamter, int iSample, int iEvent) = 0;
   virtual double ReturnKinematicParameter(double KinematicVariable, int iSample, int iEvent) = 0;
   virtual std::vector<double> ReturnKinematicParameterBinning(std::string KinematicParameter) = 0; //Returns binning for parameter Var
-  virtual const double* ReturnKinematicParameterByReference(std::string KinematicParamter, int iSample, int iEvent) = 0;
+  virtual const double* ReturnKinematicParameterByReference(std::string KinematicParamter, int iSample, int iEvent) = 0;
```
luketpickering, line 142: you know about my upset with this needless mess. But I've fixed it in a branch that we can discuss after this is merged

samplePDF/samplePDFFDBase.h
```diff
@@ -200,29 +193,28 @@ class samplePDFFDBase :  public samplePDFBase
...
-
-  /// An axis to set binned oscillation weights
-  TAxis *osc_binned_axis;
+  std::vector<OscillatorBase*> NuOscProbCalcers;
+  std::string NuOscillatorConfigFile;
+  
   //===============================================================================
   
   //Variables controlling oscillation parameters
   bool doubled_angle;
```
luketpickering, line 202: should be in the osc calculator?

samplePDF/samplePDFFDBase.h
luketpickering: The reason is that people can look up documentation for yaml-cpp and work out how to interact with the objects, if you wrap everything, either you have to supply your own wrapped interface to the library features, or you hide possibly useful features of the library.

luketpickering: Plus, it'll take probably 42 seconds with find/replace to change it

splines/splineFDBase.cpp
```diff
@@ -168,21 +171,19 @@ void splineFDBase::TransferToMonolith()
                   int iCoeff=coeffindexvec[splineindex];
                   getSplineCoeff_SepMany(splineindex, tmpXCoeffArr, tmpManyCoeffArr);
 
-                  #ifdef MULTITHREAD
-                  #pragma omp parallel for
-                  #endif
-                  for(int i=0; i<splineKnots; i++){
+                  for(int i = 0; i < splineKnots; i++){
 
                     if(tmpXCoeffArr[i]==-999){
```
luketpickering, line 176: hard coded magic numbers

splines/splineFDBase.cpp
```diff
@@ -394,13 +395,13 @@ std::vector<TAxis *> splineFDBase::FindSplineBinning(std::string FileName, std::
   TFile *File = new TFile(FileName.c_str());
```
luketpickering, line 395: best to do something like:

```c++
auto File = std::unique_ptr<TFile>(TFile::Open(Filename.c_str()));
```

splines/splineFDBase.cpp
```diff
@@ -414,9 +415,8 @@ std::vector<TAxis *> splineFDBase::FindSplineBinning(std::string FileName, std::
     Obj = File->Get("dev_tmp.0.0");
```
luketpickering, line 415: can use typed `TFile::Get<T>`

splines/splineFDBase.cpp
```diff
@@ -683,7 +682,7 @@ void splineFDBase::PrepForReweight()
       UniqueSystSplines[iSpline]->GetKnot(iKnot, xPoint, yPoint);
       if (xPoint == -999 || yPoint == -999)
```
luketpickering, line 683: magic numbers

luketpickering (approved):

splines/splineFDBase.cpp
EdAtkin: these checks on -999 don't actually make any sense so I'll remove them. The spline coefficients are initialised to -999 here: https://github.com/mach3-software/MaCh3/blob/develop/splines/splineFDBase.cpp#L767
But then we immediately call GetCoef() where if the there is some failure here we would not get -999 out. So I have removed this check.

splines/splineFDBase.cpp
EdAtkin: another case where this won't actually do anything since if GetKnot fails it won't just return -999.

splines/splineFDBase.cpp
EdAtkin: agree!

samplePDF/samplePDFFDBase.cpp
```diff
@@ -51,6 +41,168 @@ samplePDFFDBase::~samplePDFFDBase()
...
+    MACH3LOG_INFO("Adding cut on {} with bounds {} to {}", SelectionCuts["KinematicStr"].as<std::string>(), SelectionBounds.back().at(0), SelectionBounds.back().at(1));
+    SelectionVec = {KinematicParamter, low_bound, up_bound};
+    StoredSelection.push_back(SelectionVec);
+  }
+  NSelections = SelectionStr.size();
+
+  return;
+}
+
+void samplePDFFDBase::Initialise() {
```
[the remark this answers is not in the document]
EdAtkin: Yeah I don't like how clunky this is at the moment. You basically yo-yo between the base class and the derived class because the base class doesn't even know what to do into it's been told how much MC there is etc by the derived class. So it is quite backwards at the moment and stems from this slight confusion about the ownership of the underlying variables e.g. x_var is not actually owned by samplePDFFDBase it's owned by ,say, samplePDFDuneBeamFD. Probably this needs some restructuring where the simple reading of the MC is done by a separate object and then this gets passed to samplePDFDuneBeamFD and samplePDFFDBase by the constructor. I need to think about this more but I think it would clean up some of this bouncing about. I will raise this as an issue.

P.S. I don't like that this is called "Initiliasise" and that we also have an "Init" function...

samplePDF/samplePDFFDBase.cpp
EdAtkin: See above comment reply. I agree that this needs simplifying!

samplePDF/samplePDFFDBase.h
EdAtkin: Short answer is yes for now. The ND280 code has a base class for the ND which then inherits from samplePDFBase. The goal for the next major release (or 3.0.0) is the samplePDFND object (which currently just lives in the T2K repo because it is still a bit T2K-specific) will be combined with developments we've made with the far detector code.

The reason this hasn't happened is because there are large structural differences which we're slowly trying to remove. For example the FD code has this nested structure for the oscillation channels which the ND does not care about. The ND code also keeps all samples in one samplePDFND object with just an extra dimension for each sample, making all the multi-threading etc. more efficient.

The plan is to just have a single samplePDFBase class which both ND sample and FD samples will inherit from and this is definitely what we're aiming for it just needs us to remove a few structural pathologies from the FD code.

samplePDF/samplePDFFDBase.h
EdAtkin: this has been removed :)

samplePDF/samplePDFFDBase.h
EdAtkin: agree in general with this. This variable has actually been removed as it's not needed anymore.

samplePDF/samplePDFFDBase.h
EdAtkin: also removed

samplePDF/FDMCStruct.h
```diff
@@ -10,7 +9,7 @@ struct fdmc_base {
   bool signal; // true if signue                                              
   int nEvents; // how many MC events are there                              
   std::string flavourName;
-  int *isbound;
+
   int **Target; // target the interaction was on
```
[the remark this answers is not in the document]
EdAtkin: there is no good reason. It is basically how it has been on T2K since before I joined. This needs to be changed and is somewhere on the list. I will open this as an issue. I would actually prefer it if we created some small MaCh3Event object and each samplePDF kept a vector of these which should be more efficient cache hit wise.

cmake/Modules/NuOscillatorSetup.cmake
```diff
@@ -0,0 +1,92 @@
...
+IsTrue(MaCh3_LOW_MEMORY_STRUCTS_ENABLED DAN_DOUBLE)
+SwitchLogic(DAN_DOUBLE)
+
+# Get the compile options for MaCh3CompilerOptions
+get_target_property(compile_options MaCh3CompilerOptions INTERFACE_COMPILE_OPTIONS)
+
+# Join the compile options list into a space-separated string
+string(REPLACE ";" " " compile_options_string "${compile_options}")
+
+#KS this may look hacky however CPM isn't build for passing stuff like this. If CMAKE_CUDA_ARCHITECTURES is passed CPM it will be string not list. Thus we convert it to list
```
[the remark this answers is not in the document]
KSkwarczynski: Cheers, bit late but tried to resolve cmake comments here https://github.com/mach3-software/MaCh3/pull/182

PR #169, opened 2024-10-16
Draft: Picker24/feature/generic binning and return kinematic parameter tidy

luketpickering:
# Pull request description

changes ahead of new MaCh3 core merge.

## Changes or fixes
* adds new 'generic' ND binning option.
  + Currently only N independent contiguous (uniform or variable width) axes, the ND binning is the simple product of these N axes.
* simplifies ReturnKinematicParameter[ByReference] interface

```c++
  virtual const double &
  ReturnKinematicParameterByReference(int KinematicParameter, int iSample,
                                      int iEvent) = 0;
  // LP - use this to implement kinematic parameters that are not doubles (so we
  // cannot return the address of one), but can be cast to doubles, note that this
  // will likely cause extreme confusion and the fact we might need this
  // indicates an issue in our abstraction. Parameters defined in
  // implementations of this method can not be used as projection variables,
  // which are kept as pointers to doubles.
  virtual double ReturnKinematicParameter(int KinematicParameter, int iSample,
                                          int iEvent) = 0;
```

## Examples

New ND binning option
```yaml
Binning:
  Axes:
    - VarStr: "RecoNeutrinoEnergy" 
      VarBins: [0.,  0.5,  1.,  1.25, 1.5, 1.75, 2., 2.25, 2.5, 2.75, 3., 3.25, 3.5, 3.75, 4., 5., 6., 10.]
    - VarStr: "ERecQE" 
      Uniform: [10, 0, 10]
    - VarStr: "EHadRec" 
      Uniform: [10, 0, 10]
```


KSkwarczynski:
CI which has ROOT compiled with c++14 is failing:
https://github.com/mach3-software/MaCh3/actions/runs/11407213472/job/31742713856?pr=169

Which is related to bumping c++ standard.

This isn't no stopper but probably should discuss on gneral meeting if everyone is fine or if there are poeple who will be affected by this change etc

PR #172, opened 2024-10-17
Re-add get1DVarHist back into samplePDFFDBase

dbarrow257:
# Pull request description
During the previous merge, get1DVarHist was dropped. This was likely because of getting interaction mode information back from the ReturnKinematicVariable[ByReference] as it returns a double and mode was defined as an integer. Add this function back in whilst also defining the getNSamples and getNMCEvent functions.


samplePDF/FDMCStruct.h
```diff
@@ -7,7 +7,8 @@ struct fdmc_base {
   int nupdg;
   int nupdgUnosc;
   bool signal; // true if signue                                              
-  int nEvents; // how many MC events are there                              
+  int nEvents; // how many MC events are there
+  double ChannelIndex;
```
KSkwarczynski, line 11: What is `ChannelIndex`?

dbarrow257: In FD code, there is the concept of oscillation channels. Each oscillation channel has an associated FDMCStruct object. These channels' indexing depends entirely on the ordering defined in the sample YAML. 

If you want to loop over each oscillation channel and make an event rate breakdown by oscChan, you need some way to select events in oscillationChannel j (j=6, for example). To do this from ReturnKinematicParameter, we need to store the index of the oscillation channel somewhere in it's memory, thus ChannelIndex.

You can get name of the j'th oscillationChannel index from FDMCStruct.flavourName

dbarrow257: In DUNE code for example:

```
const double* samplePDFDUNEAtm::GetPointerToKinematicParameter(KinematicTypes KinPar, int iSample, int iEvent) {
  double* KinematicValue;


  switch (KinPar) {
  case kTrueNeutrinoEnergy:
    KinematicValue = &(dunemcSamples[iSample].rw_etru[iEvent]);
    break;
  case kRecoNeutrinoEnergy:
    KinematicValue = &(dunemcSamples[iSample].rw_erec[iEvent]);
    break;
  case kTrueCosZ:
    KinematicValue = &(dunemcSamples[iSample].rw_truecz[iEvent]);
    break;
  case kRecoCosZ:
    KinematicValue = &(dunemcSamples[iSample].rw_theta[iEvent]);
    break;
  case kOscChannel:
    KinematicValue = &(MCSamples[iSample].ChannelIndex);
    break;
  case kMode:
    KinematicValue = MCSamples[iSample].mode[iEvent];
    break;
  default:
    MACH3LOG_ERROR("Unknown KinPar: {}",KinPar);
  }
  
  return KinematicValue;
}

KSkwarczynski (approved):

PR #175, opened 2024-10-20
Start From Previous Chain

KSkwarczynski:
# Pull request description
Implement staring from previous chain

## Examples
blarb
image


dbarrow257 (commented):
It's looking good to me - Correct me if I'm wrong, but I believe the systematic configs are now saved to the MCMC output. Would it be possible to compare the saved configs to the ones that have been 'setup' in the exec which calls the StartFromPrevious function?

KSkwarczynski:
cheers for comment it should check if configs are identical

PR #177, opened 2024-10-22
add python to the docker images

ewanwm:
# Pull request description

Include python installation in the docker images so that pyMaCh3 can be used in them 
(I am not super experienced with writing dockerfiles so not sure if this is the best way)


KSkwarczynski:
I think all contianters should have ypthon. Definately Alma9 one has python3 so you should symlink python3 to python

PR #178, opened 2024-10-22
Auto generation python docs

ewanwm:
# Pull request description

"These will be automatically generated. It should be pretty easy to get working" - Me on Monday. Well here we 15 commits and dozens of failed attempts later. But... it works!

## Changes or fixes

Extend the documentation generation action to also build sphinx documentation for python and add it to a sub folder in the existing html which gets uploaded to github pages. 

One note is that it now takes a bit longer to generate as it needs to actually build the python module in order to make the documentation for it. In future it might be good to have it only run the sphinx generation part if the python code has been changed (this is tricky but I think not impossible to do). Or only generate new documentation only on tags rather than every PR.

### DIFF
--- a/.github/workflows/Doxygen.yml
+++ b/.github/workflows/Doxygen.yml
@@ -49,18 +49,17 @@ jobs:
 
       ## EM: In order to build the python documentation we first need to build the pyMaCh3 module
       - name: Install Python dependencies
-        uses: py-actions/py-dependency-install@v4
-        with:
-          path: "requirements.txt"
+        run: pip install -r "requirements.txt"
         
       - name: Build pyMaCh3
         run: pip install .
 
       ## EM: Now generate the sphinx documentation
-      - name: Sphinx Build              
-        uses: ammaraskar/sphinx-action@0.4
-        with:
-          docs-folder: "Doc/sphinx"
+      - name: Sphinx Build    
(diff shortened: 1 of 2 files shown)


.github/workflows/Doxygen.yml
```diff
@@ -46,11 +46,79 @@ jobs:
...
+      - uses: geekyeggo/delete-artifact@v5
+        with:
+          name: DoxygenHTML
+          
+      - name: Build MaCh3
+        run: | 
+          mkdir build
+          cd build 
+          cmake -DMaCh3_PYTHON_ENABLED=ON ../
+          make
```
KSkwarczynski, line 90: most gitbug action have 4 threads
make -j4

this will reduce github action time we use ;)

KSkwarczynski (approved):
Linter isn't mandatory, but if you have time could you make it happy?

ewanwm:
It looks like it's angry that [this line](https://github.com/mach3-software/MaCh3/blob/57fb4350dfe743362d0c62ca677731cc0695a790/Doc/mainpage.md?plain=1#L1) exists, but i think that's needed for doxygen to find the file right?

KSkwarczynski:
Ok then ignore linter...
It sometimes is helpfull

PR #183, opened 2024-10-24
Remove Garbage

KSkwarczynski:
# Pull request description
WARNING, this will break DUNE! 
Changes here are not backward compatibilte

## Changes or fixes
- We no longer need special treatment for cov osc in fitter base. Now cov osc should be added same way as any other systematics
- Stop with terrbile hardcdoing of fixing baseline etc
- stop using const char, we should do this long time ago
- Remove many getter in cov osc, we should enforce using new more gnereir getters

### DIFF
--- a/covariance/covarianceBase.cpp
+++ b/covariance/covarianceBase.cpp
@@ -1,7 +1,7 @@
 #include "covariance/covarianceBase.h"
 
 // ********************************************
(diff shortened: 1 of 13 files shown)


EdAtkin (approved):
looks good, you even kept the python version up to date!

dbarrow257 (approved):
I have no comments - already tested on DUNE and found no problems.

PR #184, opened 2024-10-25
More Flexible Start FromPrevious

KSkwarczynski:
# Pull request description
Now Start From previous is in Fitter Base this mean every algorith can use it. Hovwerv it is virtual thus every algorithm can load some specyfic inofrmation.

Cheers Dan for comment


dbarrow257 (approved):
Looks good - thanks :)

PR #188, opened 2024-10-27
Remvoe Size...

KSkwarczynski:
# Pull request description
What's the point of having size and _fNumPar ? 

This might break seveal T2K covariances but whatever...


dbarrow257 (approved):
Looks good - not for this PR, but looks like there's a lot of repeated code in the two init functions of covarianceBase. Could be good to merge them where possible.

PR #189, opened 2024-10-28
Feature tidy up

EdAtkin:
# Pull request description
Mainly tidying up some bits of code especially the FarDetectorStruct class. Removing allocation of lots of arrays on the heap to use std::vectors instead. Also a general bit of tidy up here and there.

## Changes or fixes
Rename of FDMC struct to FarDetectorStruct. Also moving many of the members of this struct to be vectors. Small clean up of splineFDBase class.

## Examples
e.g. int* modes = new int[nEvents]; -> std::vector<int> modes; modes.resize(nEvents);

### DIFF
--- a/samplePDF/FarDetectorCoreInfoStruct.h
+++ b/samplePDF/FarDetectorCoreInfoStruct.h
@@ -2,6 +2,14 @@
 
 /// @brief constructors are same for all three so put in here
 struct FarDetectorCoreInfo {
+  FarDetectorCoreInfo() : isNC{nullptr} {}
+  FarDetectorCoreInfo(FarDetectorCoreInfo const &other) = delete;
+  FarDetectorCoreInfo(FarDetectorCoreInfo &&other) = default;
+  FarDetectorCoreInfo& operator=(FarDetectorCoreInfo const &other) = delete;
+  FarDetectorCoreInfo& operator=(FarDetectorCoreInfo &&other) = delete;
+
+  ~FarDetectorCoreInfo(){ delete [] isNC; }
+
   int nutype; // 2 = numu/signue | -2 = numub | 1 = nue | -1 = nueb           
   int oscnutype;    
   int nupdg;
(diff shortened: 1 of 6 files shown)


samplePDF/FarDetectorCoreInfoStruct.h
```diff
@@ -0,0 +1,59 @@
...
+  std::vector<double> total_w;
+
+  std::vector<int> XBin;
+  std::vector<int> YBin;
+  std::vector<int> NomXBin;
+  std::vector<int> NomYBin;
+
+  bool *isNC;
+
+  // histo pdf bins
```
KSkwarczynski, line 49: Why is it raw pointer while everyhitng else is vector?

luketpickering: Read about std::vector<bool>, it doesn't behave like other std::vector objects, or just try printing the address of the first three entries of auto myvecbool = std::vector<bool>{true,true,false};

KSkwarczynski: I feel dumb for not knowing this...

In this case there should be a destructor?

samplePDF/samplePDFFDBase.cpp
```diff
@@ -785,7 +782,7 @@ void samplePDFFDBase::SetupNormParameters(){
 //A way to check whether a normalisation parameter applies to an event or not
 void samplePDFFDBase::CalcXsecNormsBins(int iSample){
 
-  fdmc_base *fdobj = &MCSamples[iSample];
+  FarDetectorCoreInfo *fdobj = &MCSamples[iSample];
```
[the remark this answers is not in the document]
luketpickering: I'd maybe do this anyway and see if you do copy/move anywhere. If you always construct in place with `emplace_back` or `resize`, then I think its fine/best to not have copy constructors/assignment

PR #190, opened 2024-10-28
Tidies up CMake, enables core libraries to build with -Wall -Wextra -…

luketpickering:
…Werror

- updates to yaml 0.8.0
- updates to CPM 0.40.2
- removes dependency on NuHepMC/CMakeModules
- sets CXX standard to 17, can probably roll it back, but there is C++17 code in here
- separates warning/error flags from other compiler flags
  - warnings/errors become a separate interface target that we link privately to library code
- all core libraries updated (hopefully minimally) to build under stringent compiler checks. Most of this is:
  * myvect.size() -> int(myvect.size())
  * (int)mylong -> int(mylong)
  * (TObj*)SomeROOTFunc() -> static_cast<TObj*>(SomeROOTFunc())
- Also changes macro types to using declarations

# Pull request description


CMakeLists.txt
```diff
@@ -50,18 +49,7 @@ include(${CMAKE_CURRENT_LIST_DIR}/cmake/Modules/MaCh3Dependencies.cmake)
...
-# KS: If C++ standard is lower than C++ standard used for ROOT compilation things will go terribly wrong
-if(DEFINED ROOT_CXX_STANDARD AND ROOT_CXX_STANDARD GREATER CMAKE_CXX_STANDARD)
-  set(CMAKE_CXX_STANDARD ${ROOT_CXX_STANDARD})
-endif()
-cmessage(STATUS "CMAKE CXX Standard: ${CMAKE_CXX_STANDARD}")
-
-# KS: ROOT changed cmake in 6.32, we should move avay from using Luke's hack, keep it for due to compatibility
-if(ROOT_CXX_STANDARD LESS 14 AND ROOT_VERSION VERSION_LESS 6.32.00)
-  cmessage(WARNING "ROOT CXX STANDARD: ${ROOT_CXX_STANDARD}")
+  set(CMAKE_CXX_STANDARD 17)
```
KSkwarczynski, line 52: can you try ussing ROOT_CXX_STANDARD?
It is in root cofnig
https://github.com/root-project/root/blob/8d5764957e4bcb9cdc0bca44f83fb2fb4fe1815d/cmake/scripts/RootUseFile.cmake.in#L12

splines/SplineCommon.h
```diff
@@ -12,9 +12,9 @@
 #endif
 
 /// KS: We store coefficients {y,b,c,d} in one array one by one, this is only to define it once rather then insert "4" all over the code
-#define _nCoeff_ 4
+constexpr inline ushort nCoeff = 4;
 /// KS: For TF1 we store at most 5 coefficients, we could make it more flexible but for now define it here to make future changes easier to track
-#define _nTF1Coeff_ 2
+constexpr inline ushort nTF1Coeff = 2;
```
KSkwarczynski, line 17: Could you revert this to define?
These two are used in GPU.
https://github.com/mach3-software/MaCh3/blob/develop/splines/gpuSplineUtils.cu

I don't know but this could cause problem on older gpu. Updating GPU is harder than gcc. GPU conianters are pain.
No prefence for name

cmake/Modules/MaCh3Dependencies.cmake
```diff
@@ -2,34 +2,17 @@
...
 
-include(ROOT)
-if(NOT TARGET ROOT::ROOT)
-  cmessage(FATAL_ERROR "MaCh3 Expected dependency target: ROOT::ROOT")
-endif()
-
-if(ROOT_VERSION VERSION_LESS 6.18.00)
-  cmessage(FATAL_ERROR "Using ROOT version smaller than 6.18.0, this may lead to unexpected results")
-endif()
+find_package(ROOT 6.18 REQUIRED)
```
KSkwarczynski, line 15: Can you also update this
https://github.com/mach3-software/MaCh3/blob/d6414032bbf683c940f2801d2662b52f4e50fb16/cmake/Templates/MaCh3Config.cmake.in#L23

CMakeLists.txt
```diff
@@ -21,10 +22,8 @@ else()
   set(CPU_ONLY TRUE)
 endif()
 
-if(CPU_ONLY)
-  project(MaCh3 VERSION ${MaCh3_VERSION} LANGUAGES CXX)
-else()
-  project(MaCh3 VERSION ${MaCh3_VERSION} LANGUAGES CXX CUDA)
+if(NOT CPU_ONLY)
+  enable_language(CUDA)
```
KSkwarczynski, line 26: Could you move it a few line above?
around
set(MaCh3_GPU_ENABLED TRUE)

KSkwarczynski (commented):
Thanks a lot for tidyng all of this mess.
I have some minor comments which bots might have not found

KSkwarczynski:
According to CI Minuit2 flag is no longer on
image

https://github.com/mach3-software/MaCh3/blob/43e4b143abb0d737eb236a5f963878007b42c0f1/cmake/Modules/MaCh3Dependencies.cmake#L36
we were reyling on ROOT_CXX_FLAGS set via NuHEP Tools

CMakeLists.txt
```diff
@@ -76,26 +64,32 @@ endif()
...
-    #-Wold-style-cast       # Warn for c-style casts
-    #-Wconversion           # Warn on type conversions that may lose data
-    #-Werror                # Treat Warnings as Errors
+    -Wstrict-aliasing=2     # Helps detect potential aliasing issues that could lead to undefined behavior.
+    -Wuseless-cast          # Warn if you perform a cast to the same type (only in GCC >= 4.8)
+    # -Wpadded               # Warn when padding is added to a structure or class for alignment
+    -Wnull-dereference     # Warn if a null dereference is detected (only in GCC >= 6.0)
+    -Wold-style-cast       # Warn for c-style casts
+    -Wconversion           # Warn on type conversions that may lose data
+    -Werror                # Treat Warnings as Errors
```
KSkwarczynski, line 84: I can't higlhinth but 
https://github.com/mach3-software/MaCh3/blob/30cb0190dab901768dc5663bc6bedbf362225641/CMakeLists.txt#L95
this problaby should use MaCh3Warnings  instead of MaCh3CompilerOptions

CMakeLists.txt
luketpickering: yep, done

splines/SplineCommon.h
luketpickering: yep, would be good to check what nvcc versions are in use and when they support features like this, but if its needed for gpu to be a macro, thats best

CMakeLists.txt
luketpickering: yep

CMakeLists.txt
luketpickering: oh, no its not, if we 'use' the RootUseFile then it sets the standard for the whole compilation, but it doesn't expose directly what standard it uses. I'll just grab the flag-stripping code from NuHepMC

CMakeLists.txt
luketpickering: yep

CMakeLists.txt
luketpickering: I've put something a bit hacky in for the moment. NuOscillator uses the nuhepmc/cmakemodules still and it ends up setting up root, so I'd need to PR it out of that to properly just let ROOT set itself up. But this builds for the moment

cmake/Modules/NuOscillatorSetup.cmake
```diff
@@ -60,9 +60,9 @@ string(REPLACE " " ";" CMAKE_CUDA_ARCHITECTURES "${CMAKE_CUDA_ARCHITECTURES}")
 #Try adding Oscillator Class
 CPMAddPackage(
   NAME NuOscillator
-    VERSION 1.0.0
-    GITHUB_REPOSITORY "dbarrow257/NuOscillator"
-    GIT_TAG "tags/v1.0.0"
+    VERSION 1.0.1
+    GITHUB_REPOSITORY "luketpickering/NuOscillator"
+    GIT_TAG "feature/root_cmake_find_package"
```
KSkwarczynski, line 65: I think this should be resolved

covariance/covarianceBase.h
```diff
@@ -274,17 +274,19 @@ class covarianceBase {
...
-    for (unsigned int i = 0; i < parsSize; i++) {
+    int parsSize = int(pars.size());
+    for (int i = 0; i < parsSize; i++) {
       fParProp_PCA(i) = pars[i];
     }
     //KS: Transfer to normal base
     TransferToParam();
   }
+  /// @brief Get number of params in normal Base, if you want something which will work with PCA as well please use getNpars()
+  inline int getSize() { return _fNumPar; }
```
KSkwarczynski, line 289: why was this resurected?

covariance/covarianceBase.h
```diff
@@ -385,6 +387,8 @@ class covarianceBase {
   /// The input root file we read in
   const std::string inputFile;
 
+  /// Total number of params, deprecated, please don't use it
+  int size;
```
KSkwarczynski, line 391: why was this resurected?

tests/manager_tests.cpp
```diff
@@ -0,0 +1,18 @@
...
+  REQUIRE(lineup["1B"].as<std::string>() == "Prince Fielder");
+
+  OverrideConfig(lineup, "1B", "Fielder formerly know as Prince");
+
+  REQUIRE(lineup["1B"].as<std::string>() == "Fielder formerly know as Prince");
+
+  OverrideConfig(lineup, "1B", 123);
+
+  REQUIRE(lineup["1B"].as<unsigned>() == 123u);
+}
```
KSkwarczynski, line 18: I know you don't agree with me but I think this test should be in https://github.com/mach3-software/MaCh3Tutorial/tree/main/CIValidations but not gonna force it

KSkwarczynski (commented):
Soma comments

tests/manager_tests.cpp
luketpickering: done

cmake/Modules/NuOscillatorSetup.cmake
luketpickering: now is

KSkwarczynski (approved):

PR #195, opened 2024-11-01
adds -DMaCh3_WERROR_ENABLED=OFF options if you just can't even with GCC

luketpickering:
# Pull request description


KSkwarczynski (approved):
grand

PR #200, opened 2024-11-04
Sample PDF Factory

KSkwarczynski:
# Pull request description
Implement SamplePDF Factory, best part is that it uses template so MaCh3 doesn't have to know about type of class. I tested and it works with Tutorial SasmplePDF

## Examples
https://github.com/mach3-software/MaCh3Tutorial/pull/37

### DIFF
--- a/covariance/covarianceBase.cpp
+++ b/covariance/covarianceBase.cpp
@@ -311,8 +311,6 @@ void covarianceBase::init(const std::vector<std::string>& YAMLFile) {
   MACH3LOG_INFO("----------------");
   MACH3LOG_INFO("Found {} systematics parameters in total", _fNumPar);
   MACH3LOG_INFO("----------------");
-
-  return;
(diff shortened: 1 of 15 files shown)


covariance/covarianceBase.cpp
```diff
@@ -592,7 +587,7 @@ void covarianceBase::proposeStep() {
 // "Randomize" the parameters in the covariance class for the proposed step
 // Used the proposal kernel and the current parameter value to set proposed step
 // Also get a new random number for the randParams
-void covarianceBase::randomize() {
+void covarianceBase::randomize() _noexcept_ {
```
henry-wallace-phys, line 590: Why are we putting this in a no-except block? That feels a bit spooky to me

henry-wallace-phys: [same with the ones below this]

PR #203, opened 2024-11-05
Caseinsentisitve param group

KSkwarczynski:
# Pull request description
Luke made comment that we should be case isnetivive for parmaeter group.
Added small fucntion to redcue of copy pasting some code related with DetId. This should make even simpler modyfign treatment in the future.


dbarrow257 (approved):
LGTM

PR #208, opened 2024-11-07
Fix py ma ch3 pip install

ewanwm:
# Pull request description

Previously when pip installing, python was unable to find the sub-modules of pyMaCh3. Have now restructured the python directory to be more similar to a "normal" python only module so that it can now find it. ( This restructure also makes it easier to add any python only code in future if that's something we ever want to do ). the module now works when installing normally and with pip.

## Changes or fixes

- restructured python directory
- update readme a bit

### DIFF
--- a/python/pyMaCh3/__init__.py
+++ b/python/pyMaCh3/__init__.py
@@ -1,0 +1,3 @@
+from ._pyMaCh3 import __doc__, fitter, manager, plotting, sample_pdf, splines
+
+__all__ = ["__doc__", "fitter", "manager", "plotting", "sample_pdf", "splines"]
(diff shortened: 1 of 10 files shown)


python/pyMaCh3/__init__.py
```diff
@@ -0,0 +1,3 @@
+from ._pyMaCh3 import __doc__, fitter, manager, plotting, sample_pdf, splines
+
+__all__ = ["__doc__", "fitter", "manager", "plotting", "sample_pdf", "splines"]
```
henry-wallace-phys, line 3: With the other fixes this can be left blank and it'll work!

ewanwm: it seems like this is still needed, but the other python module files (manager.py, fitter.py ...) aren't any more so I've removed them

KSkwarczynski (approved):

PR #209, opened 2024-11-07
Correctly ignore C-style cast in CUDA header, short int->int to avoid…

henry-wallace-phys:
… information loss in type conversion

# Pull request description
Fixes to ensure compilation with CUDA

## Changes or fixes
-> Fix typo when ignoring warnings from CUDA header
-> Fix int->short int conversion


KSkwarczynski:
You copy int into short int?
https://github.com/mach3-software/MaCh3/blob/c919f706a6824008e03c75f35321497048523a17/splines/gpuSplineUtils.cu#L61
this need fixing

dbarrow257 (commented):
Is there any reason why we can't use the short int which was used previously?

dbarrow257 (approved):
LGTM

PR #230, opened 2024-11-29
WAIC

KSkwarczynski:
# Pull request description
Implement WAIC based on Gellman paper
image
https://arxiv.org/pdf/1307.5928
## Changes or fixes
-Extend usage of vectors instead of raw pointers

## Examples
Using some dummy values  I obtained WAIC  very similiar to DIC. It make sense both methods give similair results
image

### DIFF
--- a/mcmc/SampleSummary.cpp
+++ b/mcmc/SampleSummary.cpp
@@ -44,36 +44,34 @@ SampleSummary::SampleSummary(const int n_Samples, const std::string &Filename, s
   OutputTree = nullptr;
   rnd = std::make_unique<TRandom3>();
 
-  DataHist = new TH2Poly*[nSamples];
-  DataHist_ProjectX = new TH1D*[nSamples];
-  DataHist_ProjectY = new TH1D*[nSamples];
-  NominalHist = new TH2Poly*[nSamples];
-  PosteriorHist = new TH1D**[nSamples];
-  W2NomHist = new TH2Poly*[nSamples];
-  w2Hist = new TH1D**[nSamples];
+  DataHist.resize(nSamples);
+  DataHist_ProjectX.resize(nSamples);
+  DataHist_ProjectY.resize(nSamples);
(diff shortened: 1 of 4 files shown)


mcmc/SampleSummary.cpp
```diff
@@ -2558,6 +2460,32 @@ double SampleSummary::GetModeError(TH1D* hpost){
   return Mode_Error;
 }
 
+// ****************
+void SampleSummary::StudyInformationCriterion(M3::kInfCrit Criterion) {
+// ****************
+  MACH3LOG_INFO("******************************");
+  switch(Criterion) {
+    //  TSpline3 (third order spline in ROOT)
```
clarenc3, line 2468: Does this comment really belong here?

mcmc/SampleSummary.cpp
```diff
@@ -2615,7 +2541,61 @@ void SampleSummary::StudyDIC() {
   const double DIC_stat = Dhat + 2 * p_D;
   MACH3LOG_INFO("Effective number of parameters following DIC formalism is equal to: {:.2f}", p_D);
   MACH3LOG_INFO("DIC test statistic = {:.2f}", DIC_stat);
-  MACH3LOG_INFO("******************************");
+}
+
```
clarenc3, line 2545: Since you refer to "equation 3" later on in the code, please put the reference in the header before the function declaration, or here in the implementation before the function declaration

KSkwarczynski: They are in header file this way it should be readable in doxygen
https://github.com/mach3-software/MaCh3/blob/01451f03fb9871fc15bcfded02aaf2940a6ed59e/mcmc/SampleSummary.h#L153

mcmc/SampleSummary.cpp
KSkwarczynski: no crap...

PR #237, opened 2024-12-09
Feature gpu compile options

EdAtkin:
# Pull request description
Making separate GPU and CPU compiler options. This is primarily so that CPU-only flags can be exported more easily to NuOscillator without the need to do some horrible regex comparisons.

## Changes or fixes
A separate GPU option now exists where all of the CUDA-related flags now live.


CMakeLists.txt
```diff
@@ -21,6 +21,8 @@ if(CUDAToolkit_FOUND AND NOT(USE_CPU))
   set(MaCh3_GPU_ENABLED TRUE)
   enable_language(CUDA)
   set(CPU_ONLY FALSE)
+  #Create a seperate target for GPU compiler options
+  add_library(MaCh3GPUCompilerOptions INTERFACE)
```
KSkwarczynski, line 25: can you just move it away from if GPU?

But make it empty for CPU only build? I wonder if this will not simplyfy logic

CMakeLists.txt
```diff
@@ -168,23 +170,35 @@ if(CPU_ONLY)
...
 endif()
 
 if(MaCh3_LOW_MEMORY_STRUCTS_ENABLED)
-  target_compile_definitions(MaCh3CompileDefinitions INTERFACE _LOW_MEMORY_STRUCTS_)
-endif()
+  target_compile_definitions(MaCh3CompilerOptions INTERFACE _LOW_MEMORY_STRUCTS_)
+  if(MaCh3_GPU_ENABLED)
+    # ETA - think the _LOW_MEMORY_STRUCTS_ also needs this definition
+    target_compile_definitions(MaCh3GPUCompilerOptions INTERFACE _LOW_MEMORY_STRUCTS_)
+  endif(MaCh3_GPU_ENABLED)
```
KSkwarczynski, line 187: Is it really neccessary? I think it should be fine wihtout but haven't tested

CMakeLists.txt
```diff
@@ -199,6 +213,15 @@ get_target_property(link_libraries MaCh3CompilerOptions INTERFACE_LINK_LIBRARIES
...
 
+if(MaCh3_GPU_ENABLED)
+  #KS: Retrieve and print the compile options and link libraries for maximal verbose for GPU
+  get_target_property(gpu_compile_options MaCh3GPUCompilerOptions INTERFACE_COMPILE_OPTIONS)
+  get_target_property(gpu_link_libraries MaCh3GPUCompilerOptions INTERFACE_LINK_LIBRARIES)
+  #Nice print out of the GPU options if GPU is enabled
+  cmessage(STATUS "Compile options for MaCh3GPUCompilerOptions: ${gpu_compile_options}")
+  cmessage(STATUS "Link libraries for MaCh3GPUCompilerOptions: ${gpu_link_libraries}")
+endif()
+
```
KSkwarczynski, line 224: out of curitosty can you add pic of this printout in PR desciprion?

CMakeLists.txt
```diff
@@ -168,23 +170,35 @@ if(CPU_ONLY)
...
+  set_target_properties(MaCh3GPUCompilerOptions PROPERTIES EXPORT_NAME GPUCompilerOptions)
+  #If GPU enable then install both the CPU and GPU options
+  install(TARGETS MaCh3CompilerOptions MaCh3GPUCompilerOptions MaCh3CompileDefinitions
+    EXPORT MaCh3-targets
+    LIBRARY DESTINATION lib/)
+else()
+  install(TARGETS MaCh3CompilerOptions MaCh3CompileDefinitions
     EXPORT MaCh3-targets
     LIBRARY DESTINATION lib/)
+endif()
```
KSkwarczynski, line 201: this is ugly, I think if MaCh3GPUCompilerOptions  is always apparent will make this bit nicer

CMakeLists.txt
EdAtkin: yep good point! I've changed this both in the main CMakeList.txt and also manager/CMakeLists.txt and splines/CMakeLists.txt

CMakeLists.txt
EdAtkin: yep this is fixed now

CMakeLists.txt
EdAtkin: removed this, you're right that it is not used.

KSkwarczynski (approved):

PR #253, opened 2024-12-18
tidy: Step Scale Sanity chekc et. al.

KSkwarczynski:
# Pull request description
We often use defined step scale. However we know based on literature what it should be. Now simply throw warning if step scale is very different. This should help spot mistakes.

## Changes or fixes
- No longer have to covnert const char to string in adaptive code
- Now default error is 1 not 0. error being 0 can be interpeted as fixed param and we don't want this
- remove a lot of blank line in plotting
- raw pointer -> std::vecotr in cov
- more bibliography (hope is usefull)
- use new LLH scan bot

### DIFF
--- a/covariance/covarianceBase.cpp
+++ b/covariance/covarianceBase.cpp
@@ -888,8 +888,8 @@ void covarianceBase::SetBranches(TTree &tree, bool SaveProposal) {
 // ********************************************
 void covarianceBase::setStepScale(const double scale) {
 // ********************************************
-  if(scale == 0) {
-    MACH3LOG_ERROR("You are trying so set StepScale to 0 this will not work");
+  if(scale <= 0) {
+    MACH3LOG_ERROR("You are trying so set StepScale to 0 or negative this will not work");
     throw MaCh3Exception(__FILE__ , __LINE__ );
   }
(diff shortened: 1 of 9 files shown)


covariance/covarianceBase.cpp
```diff
@@ -904,21 +888,24 @@ void covarianceBase::SetBranches(TTree &tree, bool SaveProposal) {
...
 // ********************************************
-  if(scale == 0)
-  {
+  if(scale == 0) {
     MACH3LOG_ERROR("You are trying so set StepScale to 0 this will not work");
     throw MaCh3Exception(__FILE__ , __LINE__ );
   }
   MACH3LOG_INFO("{} setStepScale() = {}", getName(), scale);
+  const double SuggestedScale = 2.38*2.38/_fNumPar;
+  if(std::fabs(scale - SuggestedScale)/SuggestedScale > 1) {
```
pjdunne, line 897: This won't warn you if you go too small right? ie scale=0 won't throw an error. Maybe make it >0.8?

KSkwarczynski: Case scale = 0 is handle separately just above this check. Atlhough we don't have any checks for negative so will update it

PR #276, opened 2025-01-08
feat: Telemetry

KSkwarczynski:
# Pull request description
This will need some love in future but this bot will make RAM distribution

image
image


KSkwarczynski:
doesn't work :(

PR #278, opened 2025-01-08
bugfix: quick patches

EdAtkin:
# Pull request description
A few small fixes to some Spline tools which were causing a compiler warning. Also a fix to some of the virtual functions in the splineFDBase which were apparently undefined due to an inline.

## Changes or fixes
- added in a pragma to ignore -Wnull-dereference
- being more careful which making sure that splines aren't null
- removing inline in some function defintions in splineFDBase

## Examples
No practical examples


KSkwarczynski (approved):
Looking good

PR #280, opened 2025-01-09
bugfix: spline binning printing and TMatrixD reading

EdAtkin:
# Pull request description
very tiny fixes to reading in of TMatrixD from root files and also 

## Changes or fixes
Change to have a static_cast<TMatrixDSym> in covarianceBase and change of a single print out for the binning info.


covariance/covarianceBase.cpp
```diff
@@ -115,9 +115,9 @@ void covarianceBase::init(std::string name, std::string file) {
     throw MaCh3Exception(__FILE__ , __LINE__ );
   }
 
-  TMatrixDSym *CovMat = infile->Get<TMatrixDSym>(name.c_str());
+  TMatrixDSym *CovMat = static_cast<TMatrixDSym*>(infile->Get(name.c_str()));
```
dbarrow257, line 118: This problem is due to the saved root object being a TMatrixD object (not Sym) and the TFile helper functions return null when the types don't match. As the core object is expected to be a Sym matrix, I think there would always have to be a cast somewhere so I think this is the simplest solution.

dbarrow257 (approved):
LGTM - we could really look at the TMatrixDSym again if we need but I think there always has to be a cast somewhere

covariance/covarianceBase.cpp
EdAtkin: yeah agreed, I just didn't want to go changing the type of CovMat. I think this is very minor and we can fix later if we need to :)

PR #281, opened 2025-01-10
feat: Safer Yaml openning

KSkwarczynski:
# Pull request description
Introduce new wrapper around `YAML::LoadFile`, whole implementation is being handled by yaml, but by having wrapper we can easily intorudce additional checks and in case of error we can pass file, line which should make debugging much easier

## Changes or fixes
- test_stat is no longer member of manager. For example we use manager for samplePDF cofnig. In this case test_stat make no sense. Now we have simply getter so all code should work fine

## Examples
plot


dbarrow257 (approved):
LGTM

PR #289, opened 2025-01-14
tidy: Smarter getting Norm bin

KSkwarczynski:
# Pull request description
There is plenty of repetition in how we select norm bins. Made simple template whcih allow to simplyfi and reduce hardcoding


EdAtkin (approved):
nice, I agree that this neatens up the Norm setting up.

PR #292, opened 2025-01-15
breaking: Alter treamtent of getting weight from func pars

KSkwarczynski:
# Pull request description
None of current experiemnt get weight from from func pars. Current implemntaiton requeprs at least one mutliplication per event. Mutliplication by 1 does nothing for physics but does impact perofmacne.

New implemation retain feasiblity to apply func weights but doesn't suffer for peromance issue.

Need this to work:
https://github.com/mach3-software/MaCh3Tutorial/pull/73


EdAtkin (approved):
Yeah seems good to me, shouldn't break anything.

PR #293, opened 2025-01-17
bugfix: Print output from MaCh3Logger fancy logging function if error is thrown

dbarrow257:
# Pull request description
Issue was found where if an error occurred in NuOscillator, the new fancy logging system wouldn't post the output to the console so the error was very difficult to solve. Now catch any runtime error and redirect stringstream buffer to console so error can printed to screen


KSkwarczynski (approved):
LGTM

PR #294, opened 2025-01-17
bugfix: Correctly read Signal FDMC variable from config

dbarrow257:
# Pull request description
Previous treatment of reading the 'signal' variable from the sample config was not being done in the core code. This meant that it was always false. Consequently (unless it was specifically filled in experiment specific code), NC signal events were being applied an oscillation weight. This will likely result in a small event rate change.

*** Relies upon https://github.com/mach3-software/MaCh3/pull/293 being merged first ***

## Changes or fixes
Rather than rely upon signal variable, just directly check the nupdg and nupdgUnosc variables are the same. Consequently can remove the signal entry from sample configs.


KSkwarczynski (approved):
LGTM

PR #311, opened 2025-01-27
tidy: is event selected

KSkwarczynski:
# Pull request description
There were 3 isEvent slected with lot of copy pasting. We only use one so remvoed other. If for some reason we still need them, then can bring them back but using some more clever code to reuse them.


dbarrow257 (approved):
LGTM - I'm not sure exactly how we ended up with 3 of these functions but I guess we'll find if any experiment uses them

PR #312, opened 2025-01-28
feat: PDF Handler

KSkwarczynski:
# Pull request description
This introduce PDF Handler which is suppsoed to handle Data,MC predictions. For now is only a struct to not have too many objects in SamplePDF. In future I think we could move many of converting arrays to TH1 into class. Making it real prediction class.

I didn't want to spend too much time as I am unsure whether people like this idea or not.

Haven't checked but could have minor positive impact on performance. I intorduced this for T2K ND280 and there it helped.

### DIFF
--- a/samplePDF/FarDetectorCoreInfoStruct.h
+++ b/samplePDF/FarDetectorCoreInfoStruct.h
@@ -3,6 +3,28 @@
 #include <string>
 #include "samplePDF/Structs.h"
 
+
+// ********************************************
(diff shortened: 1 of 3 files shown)


EdAtkin:
This is going to get a bit messy since the proposed re-name of samplePDFX was going to be sampleHandlerX I think?

KSkwarczynski:
On the contrary I think
You will have SampleHandler and PDFHandler while the latter deal wihth Data/MC/W2 predicitons

PR #313, opened 2025-01-29
tidy: More infor about threads

KSkwarczynski:
# Pull request description
Main idea is to print how many htreads are being used. Yesterday I enouctered somone using 64 threads which had terrible impact on performance. This minor thing is supposed to make finidng such mistakes easier.

## Examples
image


henry-wallace-phys (approved):
lgtm

PR #314, opened 2025-01-29
breaking: Map for kinematic param and string

KSkwarczynski:
# Pull request description
In experiment specyfic one must define map between string and enum.
In core we simply set a pointer to this map.
This allow to move plenty of code from expriemnt specyfic to Core

In addition  maps must be const this ensure they cannot be modified by mistake.
Since Core has a pointer to map then if pointer are nullptr then core will throw error if they are not initliased.

## Examples
For example see: https://github.com/mach3-software/MaCh3Tutorial/pull/78


EdAtkin:
A couple of random thoughts. Firstly, I guess we always have to have the reversed and non-reversed map because maps aren't bi-directional but maybe we could automatically make the inverted map?

Secondly, do you think it makes sense to just setup the map from a sample yaml file? It gets a bit tricky because you lose the enums and just use the map["string"] everywhere but I wonder if it's nicer? That way you can sort of "register" supported kinematic variables in the sample yaml?

EdAtkin (approved):
Yep, neatens up a lot of code!

PR #315, opened 2025-01-29
tidy: Janitorial work for SplineFD

KSkwarczynski:
# Pull request description
Code should be nicer and faster (by small margin...)

## Changes or fixes
- vector -> const vector& for speed
- Made templates to more easily clean std::vectors
- Binning wasn't properly deleted
- Find spline segment was making parraler but didn't use for, waste of resources...
- In find spline segment xArray was copied... waste of resources...
- lot of const for const corecctens...
- introduce `coeffOffset` to avoid multiplciation by 4 so many times...
- SetupSplines wasn't used so removed
- addef fmaf_t to avoid total mess in calcsplinewieght...

### DIFF
--- a/samplePDF/Structs.h
+++ b/samplePDF/Structs.h
@@ -1,8 +1,22 @@
 #pragma once
 
+// C++ includes
+#include <sstream>
+#include <fstream>
+#include <iostream>
+#include <vector>
+#include <iomanip>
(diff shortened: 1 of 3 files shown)


EdAtkin (approved):
Thanks for tidying :) makes it all neater.

PR #337, opened 2025-02-10
feat: Unify find spline segment

KSkwarczynski:
# Pull request description
Simply move almos identical find spline segment to spline base.
Minor hacks are need related with SplineSegments to support GPU splines for EventByEvent

### DIFF
--- a/samplePDF/samplePDFFDBase.cpp
+++ b/samplePDF/samplePDFFDBase.cpp
@@ -353,7 +353,7 @@ void samplePDFFDBase::reweight() {
       NuOscProbCalcers[0]->CalculateProbabilities(OscVec);
     } else {
       for (int iSample=0;iSample<int(MCSamples.size());iSample++) {
-	        NuOscProbCalcers[iSample]->CalculateProbabilities(OscVec);
+        NuOscProbCalcers[iSample]->CalculateProbabilities(OscVec);
       }
     }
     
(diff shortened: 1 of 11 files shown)


samplePDF/samplePDFFDBase.cpp
dbarrow257, line 1: Do these changes actually affect the step time in a positive way? The CalcWeightSpline function is already called from within an OMP block: https://github.com/mach3-software/MaCh3/blob/ff8c81c643b78aebbefe89dc43a659abaa7375af/samplePDF/samplePDFFDBase.cpp#L498-L548

splines/SplineBase.cpp
```diff
@@ -6,14 +6,97 @@
...
 
 
 // *****************************************
 SplineBase::~SplineBase() {
 // *****************************************
+}
+
+// *************************
+// CW: Only need to do the binary search once per parameter, not once per event!
+// Takes down the number of binary searches from 1.2M to 17, ha!
```
dbarrow257, line 22: I Imagine its more than 17 now ;)

dbarrow257 (commented):
LGTM - Just the check on performance if the SIMD for loop helps efficiency

samplePDF/samplePDFFDBase.cpp
KSkwarczynski: You mean #pragma omp simd?

This is not dependand on Parrarel?
SIMD is instuction for single thread. It allows to perform operation simultaneously, so sinlge thread will perform several multiplication at the same time.

Since we have quite a few norm/spline params then it should impact perfomance in positive way.

I am did same thing in ND280 code
https://github.com/t2k-software/MaCh3/blob/83dd81c30618d8d9fb05f21504ebe994177e47d3/samplePDF/samplePDFND.cpp#L946

and it had very positive impact on performance:
https://github.com/t2k-software/MaCh3/pull/742

dbarrow257: Ah okay, interesting. Well lets hope it gives us similar boost in performance.

dbarrow257 (approved):
LGTM

PR #346, opened 2025-02-15
feat: Special param bounds in LLHScan settings

yash-quark:
# Pull request description
Move Predictive and LLHScan settings outside the General node in the fitter yamls. Also introduce feature to set user-defined ranges for special params during LLH Scan, suggested by Dan Barrow.

## Changes or fixes

1. Group all LLHScan related settings together into the LLHScan node
2. Move LLHScan and Predictive nodes outside the General node for keeping General node minimal
3. Inside LLHScan, introduce the `ScanRanges` setting inside which you can specify a parameter and a range so that LLHScan picks up this particular range while scanning.

NOTE: Please suggest if it can be improved further 😃 

## Examples
When there are no special params specified, LLH Scan uses the default way to set the ranges, along with a logger info:
Screenshot 2025-02-15 at 09 39 33

You can set special params range as follows:
Screenshot 2025-02-15 at 09 42 31

which would result in this:
Screenshot 2025-02-15 at 09 42 55

If your set range is beyond the physical boundary of the parameter, [this](https://github.com/mach3-software/MaCh3/blob/29ef036061d971b718c56d9792f2d7decec14792/mcmc/FitterBase.cpp#L595C7-L602C8) will ensure it sets correct boundaries, for example:
Screenshot 2025-02-15 at 09 42 14

### DIFF
--- a/.github/workflows/Benchmark.yml
+++ b/.github/workflows/Benchmark.yml
@@ -27,7 +27,6 @@ jobs:
           # Clone the MaCh3Tutorial repository with the current branch
           git clone --branch main https://github.com/mach3-software/MaCh3Tutorial.git MaCh3Validations
           cd MaCh3Validations
-          git checkout feature/OrganiseLLHScanSettings
           mkdir build
           cd build
           # Configure the project with benchmark support enabled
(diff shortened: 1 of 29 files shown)


yash-quark (commented):
These changes were necessary to make CI work. Need to be reverted once tutorial PR is also merged

KSkwarczynski (approved):
LGTM

PR #355, opened 2025-02-20
tidy: Remove parts of terrible loops in SplineFD

KSkwarczynski:
# Pull request description
AddSample is now void not bool.
Simlfy some of stupid loops in SplineFD.
There is still SO MUCH MESS in this class but don't have more time right now


dbarrow257 (approved):
LGTM

PR #359, opened 2025-02-20
feat: Move FillSampleArray to splineFDBase

dbarrow257:
# Pull request description


splines/splineFDBase.cpp
```diff
@@ -4,6 +4,9 @@
 #pragma GCC diagnostic ignored "-Wuseless-cast"
 #pragma GCC diagnostic ignored "-Wfloat-conversion"
 
+#include "TROOT.h"
+#include "TKey.h"
+
```
KSkwarczynski, line 9: Can you add protection here?
https://github.com/mach3-software/MaCh3/blob/73fc16ca2c4475d58ef883cfcb2f7a809ea7121a/splines/SplineBase.h#L12C1-L23C29

KSkwarczynski (approved):
LGTM

We should update as cout is used in several places instead of logger but whatever at this point

PR #366, opened 2025-02-21
bugfix: Include support for 1D samples in samplePDFFDBase::PrintIntegral

dbarrow257:
# Pull request description


KSkwarczynski (approved):
Bit hacky but don''t have better idea myself...

PR #369, opened 2025-02-24
feat: Improve Vector Casting

KSkwarczynski:
# Pull request description
You cannot cast in c++ vector but you can write simple template function which will do the casting.

In addiditon use `Get` from Yaml helpers which has built in check for node exsiting
https://github.com/mach3-software/MaCh3/blob/f68b8e074e1c15a38218b2f076b4275fe2c349a4/manager/YamlHelper.h#L262
this allows to make loading Sample Config easier


dbarrow257 (approved):
LGTM - It would be good to advertise the YAML helper more as this is super useful and I had zero idea ;)

EdAtkin (approved):

PR #371, opened 2025-02-24
bugfix: LLHScanSkipVector yaml node group inside sigma variation function

yash-quark:
# Pull request description
This is just a small update to https://github.com/mach3-software/MaCh3/pull/346 where the LLHScanSkipVector was being searched under General node in the Sigma Variation function.


KSkwarczynski (approved):
LGTM

Issue #373, opened 2025-02-25
Improved detector ID treatment


EdAtkin:
**Description**
"DetID" in the sample config should not be used. Instead this should the a unique SampleName and we can add another node for a nice plotting title for the sample i.e. lines 1 and 19 need to be changed.

This makes it possible to wildcard include samples that a parameter should include, see below for example.

![Image](https://github.com/user-attachments/assets/371e8bbe-45dd-4190-a5b2-21bace0158f4)

**Describe the solution**
Change "DetID" to "SampleName" and add a "SampleTitle" (or similar).

**Additional context**
![Image](https://github.com/user-attachments/assets/c431922a-7395-467b-9848-ad1b25a3491f)

PR #374, opened 2025-02-25
breaking: det id fix

EdAtkin:
# Pull request description
Instead of reading in a separate field "DetID" from a sample config instead the "SampleName" is used and a separate field of "SampleTitle" has been added so that it can be used for plotting names.

This addresses this issue: https://github.com/mach3-software/MaCh3/issues/373

## Changes or fixes
Renaming of covariance class functions where I felt it was appropriate to use SampleName rather than DetId.
samplename is now SampleTitle and samplePDFFDBase::GetName is samplePDFFDBase::GetTitle
SampleDetID is now SampleName.
DetID in parameter yaml is now SampleID
Additional check in FitterBase on whether the SamplePDF object has a unique name.

## Examples
CovarianceXsec::AppliesToDetID -> CovarianceXsec::AppliesToSampleName

### DIFF
--- a/covariance/covarianceBase.cpp
+++ b/covariance/covarianceBase.cpp
@@ -1210,14 +1210,14 @@ void covarianceBase::SaveUpdatedMatrixConfig() {
 }
 
 // ********************************************
-bool covarianceBase::AppliesToDetID(const int SystIndex, const std::string& DetID) const {
+bool covarianceBase::AppliesToSampleName(const int SystIndex, const std::string& SampleName) const {
 // ********************************************
   // Empty means apply to all
   if (_fDetID[SystIndex].size() == 0) return true;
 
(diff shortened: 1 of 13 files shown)


pjdunne:
Thank you. This looks in the most part great. I's just ask if we need SampleName and SampleID terminology as they're both the same thing really. Don't mind which but could we have the parameter YAML and sample YAML call those two things the same thing.

EdAtkin:
So I will will change _fSampleID to _fSampleNames instead and also any functions where the name is something like *SampleID to *SampleName or *Sample.

In terms of the string used for plotting, I added a SampleTitle node into the sample yaml files and a function GetTitle(). I will go and replace GetName with GetTitle where it's needed.

I hope that answers both of the concerns above.

pjdunne (approved):

PR #375, opened 2025-02-26
bugfix: Correctly handle duplicate modes in GetEventSplines

dbarrow257:
# Pull request description


KSkwarczynski (approved):
Great as always

PR #379, opened 2025-03-03
feat: Set Burn-In from YAML file

BenJarg:
# Pull request description

Initialize burn in cut to 0, and set it from YAML file in `ProcessMCMC`, instead of hardcoding it to 1/5.


KSkwarczynski (changes requested):
Can you set it in such a way that it still uses 20% by default.

So many people just run it without propley diagnosing BurnIn.
I think it is dangerous to not use sensible burn in for default.

Even if 20% is adhoc in many cases it is sufficient.

Overall thanks for allowing to set it, this was something I was meaning to do for a very long time...

BenJarg:
Yeah that's reasonable. Now it'll default to 20% if there's no YAML entry,

KSkwarczynski (approved):
Thanks, looking good

PR #380, opened 2025-03-04
feat: Set Burn-In for Multiple Chains

BenJarg:
# Pull request description

Pick up burn-in from YAML for chain comparisons as well. Also revert to defaulting to 20% at init to make sure nothing breaks.


KSkwarczynski (approved):
Thanks

PR #381, opened 2025-03-06
bugfix: Fix CircularPrior logic

BenJarg:
# Pull request description

This fixes a few problems with `CircularPrior()` (used for delta_cp).

First, e.g. with delta_cp, this won't actually be cyclical with period 2pi, but with period pi. For example, we want 2.5pi -> 2.5pi - 2pi = 0.5pi. But currently, this sends 2.5pi -> -pi + 0.5pi = -0.5pi.

Second, the logic is broken and won't generalize to things besides delta_cp. The signs of the `fmod` terms will be the wrong sign if `UpBound` is less than zero, or if `LowBound` is more than zero. Also, if one of the two bounds is zero, you'll get a divide by zero error.

### DIFF
--- a/covariance/covarianceOsc.cpp
+++ b/covariance/covarianceOsc.cpp
@@ -68,9 +68,9 @@ void covarianceOsc::proposeStep() {
 void covarianceOsc::CircularPrior(const int index, const double LowBound, const double UpBound) {
 // *************************************
   if(_fPropVal[index] > UpBound) {
-    _fPropVal[index] = LowBound + std::fmod(_fPropVal[index], UpBound);
+    _fPropVal[index] = LowBound + std::fmod(_fPropVal[index] - UpBound, UpBound - LowBound);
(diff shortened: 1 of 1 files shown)


BenJarg:
priors

Here's an illustration of the mapping for my proposal ("new") and what the current ("old") is doing, just implementing the logic in a simple python script.

I can't think of an easy way to illustrate this directly from this MaCh3 code, as to sample with `proposeStep`, a proposed point is chosen based on a current point, which will make a Gaussian around your "current" point, and it's not obvious what the correct behavior is. Below, I show setting the step size to 0, randomly sampling current points from -pi to 3pi (should be effectively uniform on the interval -pi to pi), and calling `proposeStep` each time, to see where the proposed point is placed.

Proposed implementation             |  Current implementation
:-------------------------:|:-------------------------:
prior_new  |  prior_old

henry-wallace-phys (approved):
Good spot!

PR #384, opened 2025-03-13
feat: Override EqualBinning setting when not using binnedosc

nosektom:
# Pull request description
Because of the EqualBinningPerOscChannel handling, binned osc configuration and switching between binned and unbinned oscillation calculation for NuOscillator are unnecessary difficult to be set properly.

EqualBinningPerOscChannel comes from sample configuration (multiple configuration files) and serves as a "sort of an indicator" of binned oscillation weights for samplePDFFDBase  (see `samplePDFFDBase::SetupNuOscillator()`). But, the calculation type is set in NuOscillator. As a result, to effectively switch between different calculation types, one needs to edit multiple configuration files, which is rather bacwardy. I would prefer handling the calculation type from NuOscillator config directly.

Not sure whether this might be called a bug or a user experience feature:)

## Changes or fixes

I suggest checking the NuOscillator config for the calculation type and override the sample configuration. This worked for me as an instant work-around (see changes). In longer-term, I guess, keeping and checking both EqualBinningPerOscChannel and a CalculationType as defined in the NuOscillator config file would be better. Something like:

```
...  
NuOscillatorConfigFile = Get<std::string>(SampleManager->raw()["NuOsc"]["NuOscConfigFile"], __FILE__ , __LINE__);  
EqualBinningPerOscChannel = Get<bool>(SampleManager->raw()["NuOsc"]["EqualBinningPerOscChannel"], __FILE__ , __LINE__);  
NuOscillatorCalcType = YAML::LoadFile(NuOscillatorConfigFile)["General"]["CalculationType"].as<std::string>();  
...
```  

## Examples

When setting NuOscillator config to "Unbinned", samplePDF throws from `samplePDFFDBase::SetupNuOscillator()` with 

`Attempted to use equal binning per oscillation channel, but not binning has been set in the NuOscillator::Oscillator object`

if not all the samples are set to `EqualBinningPerOscChannel = false`.

If avoided from within `samplePDFFDBase::SetupNuOscillator()`, samplePDF will keep handling the oscillation weights as binned, though NuOscillator is set to "Unbinned". This leads to runtime errors.

### DIFF
--- a/samplePDF/samplePDFFDBase.cpp
+++ b/samplePDF/samplePDFFDBase.cpp
@@ -69,7 +69,10 @@ void samplePDFFDBase::ReadSampleConfig()
 
   // TN override the sample setting if not using binned oscillation
   if (EqualBinningPerOscChannel) {
-    if (YAML::LoadFile(NuOscillatorConfigFile)["General"]["CalculationType"].as<std::string>() != "Binned") EqualBinningPerOscChannel = false;
+    if (YAML::LoadFile(NuOscillatorConfigFile)["General"]["CalculationType"].as<std::string>() != "Binned") {
+      MACH3LOG_WARN("Tried using EqualBinningPerOscChannel while using Unbinned oscillation calculation, changing EqualBinningPerOscChannel to false");
+      EqualBinningPerOscChannel = false;
+    }
   }
   
   //Default TestStatistic is kPoisson
--- a/samplePDF/samplePDFFDBase.cpp
+++ b/samplePDF/samplePDFFDBase.cpp
@@ -66,6 +66,11 @@ void samplePDFFDBase::ReadSampleConfig()
   SampleDetID = Get<std::string>(SampleManager->raw()["DetID"], __FILE__ , __LINE__);
   NuOscillatorConfigFile = Get<std::string>(SampleManager->raw()["NuOsc"]["NuOscConfigFile"], __FILE__ , __LINE__);
   EqualBinningPerOscChannel = Get<bool>(SampleManager->raw()["NuOsc"]["EqualBinningPerOscChannel"], __FILE__ , __LINE__);
+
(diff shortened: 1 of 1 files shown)


KSkwarczynski:
To add some clarification 
`EqualBinningPerOscChannel ` is MaCh3 feature. As OscChannel is deeply tied to SamplePDF structure. On the other hand whether binned or unbinned is used is internal to NuOscillator.

In other words you can run binned oscialtions and have EqualBinningPerOscChannel then each channel will have different binning, which is equally valid as having same binning for all channels. The latter is faster as you make calcuation once.

Obviosly EqualBinningPerOscChannel will not work for Unbinned as each channel need it's own binning. Here binning is dictated by energy of each event in channel.

Hope it helps, feature is fine but I have one comment.

Could you add `MACH3LOG_WAR("tried using EqualBinningPerOscChannel  and unbinned changin EqualBinningPerOscChannel  to false") `or something better written? To avoid silently changing settings?

nosektom:
Thanks for extra clarification Kamil, yes you are right. As stated above I think even better than this workaround would be to keep track of the `EqualBinningPerOscChannel` and NuOscillator calculation type separately and check both where needed. I guess the calculation type might acquire also different values from Binned and Unbinned in future, or are these two exclusive for good.

Yes, thanks, I'll add the MACH3LOG_WAR.

KSkwarczynski:
I perosnally am not aware if there are plans to add something beyond `Binned `and `Unbinned`.

KSkwarczynski (approved):
Thanks, looks good

dbarrow257:
Too add to this - we now have support in NuOsc for SubSampling calculation type. This is essentially the sub-sampling averaging approach which was done in T2K+SK analysis. In principal, we oscillation channel could have different binning but in practice this isn't likely to occur.

PR #385, opened 2025-03-14
bugfix: fix the index of sample llh histograms

nosektom:
# Pull request description
I believe these histogram and LLH calculation indices are wrongly iterated over. Histograms reserved for LLHs of `samples[0]` (corresponding to the first `samples[0]->GetNsamples()` of `TotalNSamples` histograms) are being overwritten with LLHs from samples of `samples[ivs]`. We need `SampleIterator` index in here.

Also, is there some logic or reason why to iterate over `samples` and `samples[ivs]` (indices `ivs` and `is`) and not simply from 0 to `TotalNSamples` (see also at least one other loop in `RunLLHScan()`, where no `ivs` and `is` are used)? Like this:

```
for(unsigned int is = 0; is < TotalNSamples; ++is )
{
  hScanSamSplit[is]->SetBinContent(j+1, 2*sampleSplitllh[is]);
}
```

## Changes or fixes
Insert `SampleIterator` index instead of `is` , which runs again from 0 for each `samples[ivs]`.

### DIFF
--- a/mcmc/FitterBase.cpp
+++ b/mcmc/FitterBase.cpp
@@ -728,7 +728,7 @@ void FitterBase::RunLLHScan() {
           {
             for(int is = 0; is < samples[ivs]->GetNsamples(); ++is)
             {
-              hScanSamSplit[is]->SetBinContent(j+1, 2*sampleSplitllh[is]);
+              hScanSamSplit[SampleIterator]->SetBinContent(j+1, 2*sampleSplitllh[SampleIterator]);
               SampleIterator++;
             }
           }


nosektom:
The CI tests are failing because of the build fail due to updates related to `SampleDetID` handling...

KSkwarczynski (approved):
Thanks for fixing this.

To answer questions about iteration. FD samplePDF object has only 1 sample, while ND280 has all samples.

I think fixing CI is on @EdAtkin and @pjdunne

I will merge it now as there is nothing controversial

nosektom:
@KSkwarczynski There is certainly some non-integrity in treating the samples of samplePDFs objects since FD samplePDF actually "has" samples, but those are oscchannel samples I think. (FD samplePDF objects have -- for T2K -- 6 samples allocated at least)

KSkwarczynski:
@nosektom yes this is correct, LLH scan by sample make no sense for SK sadly at the moment

PR #392, opened 2025-03-18
feat: SampleName[Sample] might not be initialized

nosektom:
# Pull request description
There is an inherent problem here, since the `SampleName[Sample]` might not be initialized, there is no default SampleName[] in the samplePDF constructor. This leads to runtime errors when trying to do tricks and  non-standard gymnastics with the samplePDF objects.

## Changes or fixes
What I propose here is to check whether the `SampleName` vector actually has `Sample` index at all. If not (the name was not initialized), assign some default names to the samples (from where it ends up to whatever was called). Might be later reassigned to some other name, no problem with that.

Of course I am open to other solutions, like initializing the names somewhere else, but I am not really sure at what point the number of samples in the samplePDF is settled for good. Let me know.

Also, for much better user experience, we should make sure there are no samples with identical names (rename those), as any root data saving will overwrite the samples with the same name.

## Examples
This came to me when I set `LLHScanBySample: true` but used only FD samplePDFs. It went down with an error, because the names of FD samplePDFs osc channel samples were never really initialized.

Also this might be an extra issue of `FitterBase.cxx`, as `RunLLHScan` will iterate over samples of FD samplePDFs (osc channel samples) with `LLHScanBySample: true`, which I guess was meant only to run over ND samples.

### DIFF
--- a/samplePDF/samplePDFBase.cpp
+++ b/samplePDF/samplePDFBase.cpp
@@ -256,13 +256,13 @@ std::string samplePDFBase::GetSampleName(int Sample) {
 // ***************************************************************************
   if(Sample > nSamples)
   {
-    MACH3LOG_ERROR("You are asking for sample {} of the samplePDF {}. It has only {} samples!", Sample, GetName(), nSamples);
+    MACH3LOG_ERROR("You are asking for sample {} of the samplePDF {}. It has only {} samples!", Sample, GetTitle(), nSamples);
     throw MaCh3Exception(__FILE__ , __LINE__ );
   }
 
   if(SampleName.size() <= Sample)
   {
-    MACH3LOG_ERROR("The name for sample {} of the samplePDF {} was not initialized! Check whether you have set all the sample names correctly first!", Sample, GetName());
+    MACH3LOG_ERROR("The name for sample {} of the samplePDF {} was not initialized! Check whether you have set all the sample names correctly first!", Sample, GetTitle());
     throw MaCh3Exception(__FILE__, __LINE__);
   }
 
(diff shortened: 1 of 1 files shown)


KSkwarczynski (approved):
As we discussed `LLHScanBySample `at the moment is not meant to be used for SK.
However I fully understand frustration of code crashing for no reason.

I think this is fine as temporary solution. Long term goal we plan to change structure in FD when osc channel is called sample.

nosektom:
I understand the intentions, but my point is more general, I guess not "all" the ND samples (or any thinkable future samples of any kind) need to have a name at the moment the `GetSampleName(Sample)` is called. I cannot see where this is enforced.

nosektom:
If the `SampleName[Sample]` is properly initialized somewhere else before setting up the systematics, then it definitely should work. This is only for the cases, when SampleName[] was not initialized in the first place (but later might be set to something else). Not sure how to design a case test for this. Would you link me to where the ND `SampleName[]` are filled originally? Thanks

nosektom:
Also, I guess you would not let your systematic to act on something called `*_NoNameSample_*` in your config, if you are worried about https://github.com/mach3-software/MaCh3/blob/058ad645a092004ccc93ac92e5defe89d4d68e38/covariance/covarianceBase.cpp#L1213

EdAtkin:
I think for now this is ok. ND280 is the only class which uses the std::vector<std::string> SampleNames member. Everything else will use the SampleName member which is derived from samplePDFFDBase. This isn't ideal but I think for now we just move ahead.

I think the above check to make sure that it isn't completely empty is ok but also can we just consider it throwing if the name is empty? It might be more obviously that you need to go and supply the sample names this way?

nosektom:
Alright, throwing sounds also OK.

Although, throwing from a `GetSampleName()` method seems like an overkill or rather user-kill and it might not provide enough hints on what's going on. I guess, we might rather want to throw from constructor and initialization methods of the derived classes. Or enforce them to throw if not initializing the names properly.

Anyway, I don't see any easy way how to do that at the moment, so I am going with throw here... as Ed suggested.

nosektom:
Also edited the message for the already included throw to be more "informative"

PR #396, opened 2025-04-01
feat: functional parameters

hank-hua:
# Pull request description
Added an implementation of functional parameter that's hopefully faster and more flexible. 

## Changes or fixes
 - Added a new struct `FuncPars` in Structs.h, similar to spline and norm. 
   - This struct will take a valuePtr and a funcPtr. The former is set by `covarianceXsec::GetFuncPars` and the latter is set by `samplePDFFDBase::SetupFunctionalParameters`. 
 - In samplePDFFDBase, added `applyShift`, `SetupFunctionalParameters`, `RegisterIndividualFuncPar`.
   -  `RegisterFunctionalParameters` and `resetShift` are virtual so they need to be implemented in inherited classes. 

## Examples
A demo can be found in the [MaCh3 tutorial](https://github.com/mach3-software/MaCh3Tutorial/tree/hhua/demo/FunctionalParams) by running:
```
build/bin/KinemDistributionTutorial ./TutorialConfigs/FitterConfig.yaml
```

### DIFF
--- a/covariance/covarianceXsec.cpp
+++ b/covariance/covarianceXsec.cpp
@@ -34,6 +34,7 @@ void covarianceXsec::InitXsecFromConfig() {
   //KS: We know at most how params we expect so reserve memory for max possible params. Later we will shrink to size to not waste memory. Reserving means slightly faster loading and possible less memory fragmentation.
   NormParams.reserve(_fNumPar);
   SplineParams.reserve(_fNumPar);
+  FuncParams.reserve(_fNumPar);
 
   int i = 0;
   unsigned int ParamCounter[SystType::kSystTypes] = {0};
@@ -69,6 +70,7 @@ void covarianceXsec::InitXsecFromConfig() {
       ParamCounter[SystType::kNorm]++;
     } else if(param["Systematic"]["Type"].as<std::string>() == SystType_ToString(SystType::kFunc)){
       _fParamType[i] = SystType::kFunc;
+      FuncParams.push_back(GetFuncPars(param["Systematic"], i));
       _fSystToGlobalSystIndexMap[SystType::kFunc].insert(std::make_pair(ParamCounter[SystType::kFunc], i));
       ParamCounter[SystType::kFunc]++;
     } else{
@@ -93,6 +95,7 @@ void covarianceXsec::InitXsecFromConfig() {
   //KS We resized them above to all params to fight memory fragmentation, now let's resize to fit only allocated memory to save RAM
   NormParams.shrink_to_fit();
   SplineParams.shrink_to_fit();
+  FuncParams.shrink_to_fit();
 }
 
 // ********************************************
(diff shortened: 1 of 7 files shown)


samplePDF/samplePDFFDBase.cpp
```diff
@@ -673,6 +675,94 @@ void samplePDFFDBase::ResetHistograms() {
...
+    FuncPars *fp = funcParsMap[static_cast<std::size_t>(fpEnum)];
+    // if (fp->funcPtr) {
+    //   (*fp->funcPtr)(fp->valuePtr, iSample, iEvent);
+    // } else {
+    //   MACH3LOG_ERROR("Functional parameter function pointer for {} is null for event {} in sample {}", fp->name, iEvent, iSample);
+    //   throw MaCh3Exception(__FILE__, __LINE__);
+    // }
+    (*fp->funcPtr)(fp->valuePtr, iSample, iEvent);
+  }
+}
```
KSkwarczynski, line 762: Have you checked how fast is this part?

samplePDF/samplePDFFDBase.cpp
```diff
@@ -673,6 +675,94 @@ void samplePDFFDBase::ResetHistograms() {
   }
 } // end function
 
+void samplePDFFDBase::RegisterIndividualFuncPar(std::string fpName, int fpEnum, FuncParFuncType fpFunc){
+  funcParsNamesMap[fpName] = fpEnum;
+  funcParsNamesVec.push_back(fpName);
+  funcParsFuncMap[fpEnum] = fpFunc;
+}
```
KSkwarczynski, line 682: This need protection to not add same param twice.

samplePDF/Structs.h
```diff
@@ -150,6 +150,47 @@ struct XsecNorms4 {
...
+
+  /// Parameter number of this functional in current systematic model
+  int index;
+
+  /// Parameter value pointer
+  const double* valuePtr;
+
+  /// Function pointer
+  FuncParFuncType* funcPtr;
+};
```
KSkwarczynski, line 192: I think now or later we have to make decsision and split fucnitonal params into Fucntional**Shifts** and Functional**Weights**.
Becasue your code only covers fucntional shifts.

samplePDF/samplePDFFDBase.cpp
```diff
@@ -673,6 +675,94 @@ void samplePDFFDBase::ResetHistograms() {
...
+
+void samplePDFFDBase::SetupFunctionalParameters() {
+  MACH3LOG_INFO("Setting up functional parameters");
+  funcParsVec = XsecCov->GetFuncParsFromSampleName(SampleName);
+  RegisterFunctionalParameters();
+  funcParsMap.resize(funcParsNamesMap.size());
```
KSkwarczynski, line 688: RegisterFunctionalParameters is pure virtual. There has to be protection if people proeprly implemented in experiemtn specyfic code. Easiest you can imagine is throwing error if `funcParsNamesMap.size() == 0`

KSkwarczynski (changes requested):
Thanks I left several mostly minor comment but I like the improvment.

KSkwarczynski:
It would be good to also add more verbose here:
https://github.com/mach3-software/MaCh3/blob/058ad645a092004ccc93ac92e5defe89d4d68e38/covariance/covarianceXsec.cpp#L493C1-L507C2

samplePDF/samplePDFFDBase.cpp
hank-hua: I did some profiling but I'm not sure how informative the results are. In one of the earlier commits I had two unordered_maps in this for loop, which was quite slow (callgrind+kcachegrind say 20% of the cycles for each map, totaling 40%, the rest are individual function calls). So I changed to using just one vector and saving `funcPtr` inside `FuncPars`. This became a lot faster (now in `applyShifts` nearly all of the cycles are for individual function calls). 
With the same set of normalisation and spline parameters, the step times are:
- 0.04s/step (no func. pars.)
- 0.28s/step (func. pars. using unordered_maps)
- 0.15s/step (func. pars. implementation shown here).

covariance/covarianceXsec.cpp
```diff
@@ -262,6 +265,61 @@ XsecSplines1 covarianceXsec::GetXsecSpline(const YAML::Node& param) {
...
+        throw MaCh3Exception(__FILE__, __LINE__);
+      }
+    }//KinVar_i
+    func.KinematicVarStr = TempKinematicStrings;
+    func.Selection = TempKinematicBounds;
+  }
+  func.index = Index;
+  func.valuePtr = retPointer(Index);
+  return func;
+}
```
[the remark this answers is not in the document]
hank-hua: So having some sort of base class that all func/norm/spline parameters inherit? I think @EdAtkin mentioned this a while ago.

samplePDF/samplePDFFDBase.cpp
KSkwarczynski: Thanks for checking this

covariance/covarianceXsec.cpp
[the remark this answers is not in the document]
KSkwarczynski: That's probably best way forward. I don't think doing this is neccesary for this PR. But we should certainyl think about it

KSkwarczynski (commented):
# :warning: **Performance Alert** :warning:

Possible performance regression was detected for benchmark.
Benchmark result of this commit is worse than the previous benchmark result exceeding threshold `2`.

| Benchmark suite | Current: 9f85b7c2aee5a4325a0f8bc0a372acdc7398fe8f | Previous: d14efd2a9c96fc71d0c1eb92264b7ce4d27fd33b | Ratio |
|-|-|-|-|
| `Beam Reweight` | `7.29361` ms (`± 5.4122`) | `2.30502` ms (`± 380.38`) | `3.16` |

This comment was automatically generated by [workflow](https://github.com/mach3-software/MaCh3/actions?query=workflow%3AMaCh3%20Benchmark) using [github-action-benchmark](https://github.com/marketplace/actions/continuous-benchmark).

KSkwarczynski (approved):
LGTM

PR #398, opened 2025-04-08
feat: Particle-level histograms

Jude412:
# Pull request description
Declared two virtual functions in samplePDFFDBase.h for plotting particle-level 1D and 2D histograms (ie. one histogram entry for each particle). 

## Changes or fixes
The functions added are:
- virtual TH1* get1DParticleVarHist
- virtual TH2* get2DParticleVarHist

## Examples
If it is to be used, the function can be overridden in a derived class, see for example samplePDFDUNEBeamNDGAr.cpp in MaCh3_DUNE in branch jmartin/feature/NDGAr

### DIFF
--- a/samplePDF/samplePDFFDBase.h
+++ b/samplePDF/samplePDFFDBase.h
@@ -79,6 +79,9 @@ public:
   TH1* get1DVarHist(std::string ProjectionVar, std::vector< std::vector<double> > SelectionVec = std::vector< std::vector<double> >(), int WeightStyle=0, TAxis* Axis=nullptr);
   TH2* get2DVarHist(std::string ProjectionVarX, std::string ProjectionVarY, std::vector< std::vector<double> > SelectionVec = std::vector< std::vector<double> >(), int WeightStyle=0, TAxis* AxisX=nullptr, TAxis* AxisY=nullptr);
 
+	virtual TH1* get1DParticleVarHist(std::string ProjectionVarX, std::vector< std::vector<double> > SelectionVec = std::vector< std::vector<double> >(), int WeightStyle=0, TAxis* AxisX=nullptr){(void)ProjectionVarX; (void)SelectionVec; (void)WeightStyle; (void)AxisX; return nullptr;};
+	virtual TH2* get2DParticleVarHist(std::string ProjectionVarX, std::string ProjectionVarY, std::vector< std::vector<double> > SelectionVec = std::vector< std::vector<double> >(), int WeightStyle=0, TAxis* AxisX=nullptr, TAxis* AxisY=nullptr){(void)ProjectionVarX; (void)ProjectionVarY; (void)SelectionVec; (void)WeightStyle; (void)AxisX; (void)AxisY; return nullptr;};
+
   TH1* get1DVarHistByModeAndChannel(std::string ProjectionVar_Str, int kModeToFill=-1, int kChannelToFill=-1, int WeightStyle=0, TAxis* Axis=nullptr);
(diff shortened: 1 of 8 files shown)


KSkwarczynski:
Can you maybe show example in jmartin/feature/NDGAr how overided fucntion would look like?

As a noneDUNE person I am sply wondering if this is functionality that we want in core. I don't mean virtual function but whole implmentation to live in Core.

samplePDF/samplePDFFDBase.h
```diff
@@ -79,6 +79,9 @@ class samplePDFFDBase :  public samplePDFBase
   TH1* get1DVarHist(std::string ProjectionVar, std::vector< std::vector<double> > SelectionVec = std::vector< std::vector<double> >(), int WeightStyle=0, TAxis* Axis=nullptr);
   TH2* get2DVarHist(std::string ProjectionVarX, std::string ProjectionVarY, std::vector< std::vector<double> > SelectionVec = std::vector< std::vector<double> >(), int WeightStyle=0, TAxis* AxisX=nullptr, TAxis* AxisY=nullptr);
 
+	virtual TH1* get1DParticleVarHist(std::string ProjectionVarX, std::vector< std::vector<double> > SelectionVec = std::vector< std::vector<double> >(), int WeightStyle=0, TAxis* AxisX=nullptr){(void)ProjectionVarX; (void)SelectionVec; (void)WeightStyle; (void)AxisX; return nullptr;};
+	virtual TH2* get2DParticleVarHist(std::string ProjectionVarX, std::string ProjectionVarY, std::vector< std::vector<double> > SelectionVec = std::vector< std::vector<double> >(), int WeightStyle=0, TAxis* AxisX=nullptr, TAxis* AxisY=nullptr){(void)ProjectionVarX; (void)ProjectionVarY; (void)SelectionVec; (void)WeightStyle; (void)AxisX; (void)AxisY; return nullptr;};
+
```
KSkwarczynski, line 84: I know we are not doing great job but some doxygen comments would be helpfull

Jude412:
> Can you maybe show example in jmartin/feature/NDGAr how overided fucntion would look like?
> 
> As a noneDUNE person I am sply wondering if this is functionality that we want in core. I don't mean virtual function but whole implmentation to live in Core.

Hi Kamil, sure, here is what the 2D version looks like in my branch:
https://github.com/DUNE/MaCh3_DUNE/blob/jmartin/feature/NDGAr/samplePDFDUNE/samplePDFDUNEBeamNDGAr.cpp#L1021

I do think this could be put into core with some adjustments, there are just a couple of things I would need to work out. The way particle-level variables are currently implemented is as vectors in dunemc_base with as many entries as there are particles in the sample ( https://github.com/DUNE/MaCh3_DUNE/blob/jmartin/feature/NDGAr/samplePDFDUNE/StructsDUNE.h#L96 ). One variable the new function uses is the number of particles in the sample (in order to loop through them all and fill the histo). This would not be readily available in core. Currently, I have a variable counting this but this could be replaced by simply looping through until the end of the vector is reached. 

The more difficult aspect is that the function must know which event each particle belongs to. This is necessary for plotting a particle-level histogram with event-level kinematic cuts (eg. particle momentum distribution, but only for CC events). At the moment, I have a particle-level variable which just stores the event index of each particle (https://github.com/DUNE/MaCh3_DUNE/blob/jmartin/feature/NDGAr/samplePDFDUNE/samplePDFDUNEBeamNDGAr.cpp#L1065). This is something that wouldn't be readily available in core and might need a more creative fix, but I would welcome any ideas.

EdAtkin:
I think we have found a work around for this for now and we don't need to make changes to core. I have opened an issue [here](https://github.com/mach3-software/MaCh3/issues/399).

PR #402, opened 2025-04-14
feat: N Threas Sanity Check

KSkwarczynski:
# Pull request description
Do not allow people to use more than 16 threads. With so many code will be slower. Now there are plenty of modern CPU which can have even 64 or more threads.

In addition tidy yaml usage

### DIFF
--- a/manager/Monitor.cpp
+++ b/manager/Monitor.cpp
@@ -112,8 +112,27 @@ void GetCPUInfo() {
   MACH3LOG_INFO("{}", TerminalToString("lscpu | grep -m 1 -E '^CPU(:|\\(s\\)):?\\s+[0-9]+'"));
   MACH3LOG_INFO("With available threads                {}", M3::GetNThreads());
 
+  NThreadsSanity();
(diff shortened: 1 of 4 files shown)


EdAtkin (commented):
Is this always true? Like if you have 10 times the amount of MC that T2K does is using 16 threads actually a slow down?

KSkwarczynski:
I haven't done very detailed tests, even if you increase MC by 10 you will still have same number of splines, covariance operations and oscilations (assuming binned are used).

I rember at one point doing tests with SK code and seeing flatening/worsening around 10 threads.

Since this is so system dpednant I very arbitrary set it to 16.

I can bump limit to 20 or more. There will never be one good number as each experiemnt/cluster is different.

However I stronlgy think we need some sanity check that user aren't using all avaialbe CPU

EdAtkin:
fair enough. The message is very clear and easy to fix in case for whatever reason somebody wants to run with a large number of threads.

EdAtkin (approved):

PR #403, opened 2025-04-14
bugfix: Properly select test statistic

KSkwarczynski:
# Pull request description
Previously test statictic was being set via
`fTestStatistic = static_cast<TestStatistic>(SampleManager->raw()["TestStatistic"].as<int>());
`
which was simply wrong... this didn't work.
The only reason why it worked is becasue this yaml node was not set so it was defaulting to Poisson.
Now selection should work.

In addition add option to not update W2 based on T2K expierences.

### DIFF
--- a/samplePDF/samplePDFFDBase.cpp
+++ b/samplePDF/samplePDFFDBase.cpp
@@ -411,9 +411,9 @@ void samplePDFFDBase::fillArray() {
         continue;
       } 
 
-      double splineweight = 1.0;
-      double normweight = 1.0;
-      double totalweight = 1.0;
+      M3::float_t splineweight = 1.0;
+      M3::float_t normweight = 1.0;
+      M3::float_t totalweight = 1.0;
       
       if(SplineHandler){
         splineweight = CalcWeightSpline(iSample, iEvent);
--- a/samplePDF/samplePDFFDBase.cpp
+++ b/samplePDF/samplePDFFDBase.cpp
@@ -35,6 +35,10 @@ samplePDFFDBase::samplePDFFDBase(std::string ConfigFileName, covarianceXsec* xse
   samplePDFFD_data = nullptr;
   SampleName = "";
   SampleManager = std::unique_ptr<manager>(new manager(ConfigFileName.c_str()));
+
+  // Variables related to MC stat
+  FirstTimeW2 = true;
+  UpdateW2 = false;
 }
 
 samplePDFFDBase::~samplePDFFDBase()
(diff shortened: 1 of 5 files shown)


samplePDF/samplePDFFDBase.cpp
```diff
@@ -484,18 +483,20 @@ void samplePDFFDBase::fillArray() {
       //DB Fill relevant part of thread array
       if (XBinToFill != -1 && YBinToFill != -1) {
         samplePDFFD_array[YBinToFill][XBinToFill] += totalweight;
-        samplePDFFD_array_w2[YBinToFill][XBinToFill] += totalweight*totalweight;
+        if (FirstTimeW2) samplePDFFD_array_w2[YBinToFill][XBinToFill] += totalweight*totalweight;
```
EdAtkin, line 486: I'm not being helpful here but adding in an if statement creates another branch in the code. Does this not slow things down by a noticeable amount?

samplePDF/samplePDFFDBase.cpp
```diff
@@ -669,7 +672,11 @@ void samplePDFFDBase::ResetHistograms() {
   size_t nYBins = int(YBinEdges.size()-1);
   
   //DB Reset values stored in PDF array to 0.
+  // Don't openMP this; no significant gain
   for (size_t yBin = 0; yBin < nYBins; ++yBin) {
+    #ifdef MULTITHREAD
```
EdAtkin, line 677: just to check, there is not a significant gain multi-threading the outer loop but the inner loop there is a speedup?

EdAtkin (commented):
I think this looks ok just leaving some minor questions

samplePDF/samplePDFFDBase.cpp
KSkwarczynski: this is nonMP code so we don't care about speed.

For analogous code in MP loop, branch is created outside of event loop.
So branching is proportional to number of bins, not number of events so should not be significant

samplePDF/samplePDFFDBase.cpp
KSkwarczynski: Multithreading introudces a slowdown becasue operation is much faster compared with threads operations. I tested it with ND280 some time agoi where we have MUCH more bins so even if I saw slowdown there then in FD would be even worse.

I have added simd which is complety different from actual multihreading.

samplePDF/samplePDFFDBase.cpp
EdAtkin: ok great, thanks for clarifying :+1: so as long as each bin has some reasonable number of events in then this should not be a significant slow down.

PR #405, opened 2025-04-15
breaking: adaptive mcmc update

hank-hua:
# Pull request description
PR for my adaptive MCMC changes before Ed's renaming update. 

## Changes or fixes
- Added feature where adaptive matrix gets saved every time it gets changed
- Added a yaml option to set LLHScan sigma
- Added a check for `SetupNuOscillator()` so it only gets called when `osccov` is not null
- Added `-march=native` to optimise for cpu architecture, and added `-mno-avx` for debugging when using valgrind (callgrind). 
- Fixed a minor bug in `covarianceBase::updateAdaptiveCovariance`


mcmc/mcmc.cpp
```diff
@@ -140,9 +140,36 @@ void mcmc::ProposeStep() {
     syst_llh[s] = systematics[s]->GetLikelihood();
     llh += syst_llh[s];
 
-    #ifdef DEBUG
-    if (debug) debugFile << "LLH after " << systematics[s]->getName() << " " << llh << std::endl;
-    #endif
```
KSkwarczynski, line 145: I know this isn't used but why remove?

mcmc/FitterBase.h
```diff
@@ -120,6 +120,8 @@ class FitterBase {
   TFile *outputFile;
   /// Output cov folder
   TDirectory *CovFolder;
+  /// Output adaptive folder
+  TDirectory *AdaptiveFolder;
```
KSkwarczynski, line 124: I would personally use CovFolder for adaptive information.

Adaptive is related to covariance so I think it make sense conceptually to use same TDirectory

mcmc/mcmc.cpp
```diff
@@ -140,9 +140,36 @@ void mcmc::ProposeStep() {
...
+        // HH: We need to flush the file to make sure the data is written  
+        AdaptiveFolder->GetFile()->Flush();
+        std::cout << "Saving adaptive throw matrix at step " << total_steps_str << " for systematic " << systematics[s]->getName() << std::endl;
+        TMatrixDSym *throwMatrix = systematics[s]->getThrowMatrix();
+        for (int i = 0; i < throwMatrix->GetNrows(); ++i) {
+          std::cout << (*throwMatrix)(i, i) << " ";
+        }
+        std::cout << std::endl;
+      }
+    }
```
KSkwarczynski, line 172: I don't think I understand but this will save matrix not once but multiple time.
Do we really need to save adaptive matrix every time it is updated?

We lacked funcitonality for saving adaptive matrix so thanks for this but I would really think if we need to do it everytime is updated.

Also please use logger

KSkwarczynski (changes requested):

mcmc/mcmc.cpp
EdAtkin: added this back

mcmc/FitterBase.h
EdAtkin: yep I agree. I've changed this to write to the covFolder instead

mcmc/mcmc.cpp
EdAtkin: I've added in the option to now save only N iterations where this is set via the config settings. The default is 1 but if you only want to save say every 10th iteration then this should now be possible.

EdAtkin: I guess we need to decide if this lives in MCMC or not... I think it could go into the covariance class

KSkwarczynski: I forgot but already exist way to save adaptive to file
https://github.com/mach3-software/MaCh3/blob/d3a6497da7f96f4842e4a027d04bfd855d0d4688/covariance/AdaptiveMCMCHandler.cpp#L121C1-L141C1

So would be good to use it

EdAtkin: aha so you'd prefer it if it doesn't get saved in the MCMC output file put in a different file entirely? This is definitely possible and means we can remove the saving from the mcmc class

EdAtkin: this is now fixed and seems to be working with the tutorial

mcmc/mcmc.cpp
```diff
@@ -1,4 +1,5 @@
 #include "mcmc.h"
+#include "manager/MaCh3Logger.h"
```
KSkwarczynski, line 2: Super nitpicky but do we actually nedd this here?
I am certain it is included via some other header

covariance/AdaptiveMCMCHandler.cpp
```diff
@@ -1,4 +1,6 @@
 #include "covariance/AdaptiveMCMCHandler.h"
+#include "manager/YamlHelper.h"
```
KSkwarczynski, line 2: same here about Yaml.h

mcmc/MaCh3Factory.h
```diff
@@ -1,6 +1,7 @@
 #pragma once
 
 // MaCh3 includes
+#include "manager/YamlHelper.h"
```
KSkwarczynski, line 4: and here

covariance/AdaptiveMCMCHandler.h
```diff
@@ -76,6 +76,16 @@ class AdaptiveMCMCHandler{
   /// Steps between changing throw matrix
   int adaptive_update_step;
 
+  /// If you don't want to save every adpation then
+  /// you can specify this here
+  int adaptive_save_n_iterations;
+
+  /// Count the number of adaptive iterations performed
+  int iteration_counter;
```
henry-wallace-phys, line 84: Putting here but I'm not totally sure why this needs to be a separate variable

henry-wallace-phys (changes requested):
Mostly looks fine, just not sure if we need a separate header defined attribute for counting iterations given it's all done in the same method

covariance/AdaptiveMCMCHandler.h
EdAtkin: ah sorry sorry you do need this. This gets updated every iteration. I guess though this isn't required as you can calculate it from other variables? I think I can just remove it.

henry-wallace-phys (approved):
looks good

KSkwarczynski (approved):
If dr Wallace is fine then I am fine as well :)

covariance/AdaptiveMCMCHandler.h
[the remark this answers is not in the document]
EdAtkin: no this is me multi-tasking and forgetting. I've actually removed it now.

PR #408, opened 2025-04-24
feat: Yaml Merging

KSkwarczynski:
# Pull request description
Give ability to merge yaml configs, in near future I want to have ability of overwriting configs.
For example you have base T2K config, and ATM config with additional settings. You could pass both and MaCh3 would simply add/overwrite settings. For valdiations see: https://github.com/mach3-software/MaCh3Tutorial/pull/110

Also add new constuctor for manager to make easier to start from previous chain (changes will have to be done on exe level)


EdAtkin (approved):
nice feature to have!

PR #411, opened 2025-04-24
NuOsc Equal Binning Logic Change

dbarrow257:
# Pull request description
As there are many different implementations in NuOsc and only Unbinne…d will need the EqualBinning=false, change logic so Subsampling can be used with EqualBinning=true


EdAtkin (approved):

KSkwarczynski (approved):
Thanks, 
one could consider removing `YAML::LoadFile( `becasue I think it is not needed but fine as it is

PR #412, opened 2025-04-24
feat: Ability to speicfy "empty" bounds

KSkwarczynski:
# Pull request description
This is meant to help having to specify bounds like
```
SelectionCuts:
  - KinematicStr: "RecoLeptonMomentum"
    Bounds: [100., 1000000.]
```
 and instead use ""
 like here:
 SelectionCuts:
  - KinematicStr: "RecoLeptonMomentum"
    Bounds: [100., ""] 

In addition found std::vector< std::vector<double> > SelectionBounds; is not used at all so remvoed it. Did check both T2K and DUNE and neither uses it.

## Examples
test with wrong size
test

test with wrong string
test2

test with upper bound
test3

test with both bounds (pointless but why not...)
test4

Test that everything works as it used to be
image

### DIFF
--- a/samplePDF/samplePDFFDBase.h
+++ b/samplePDF/samplePDFFDBase.h
@@ -285,9 +285,7 @@ public:
   /// this is of length 3: 0th index is the value, 1st is lower bound, 2nd is upper bound
   std::vector< std::vector<double> > StoredSelection;
   /// @brief the strings grabbed from the sample config specifying the selections
-  std::vector< std::string > SelectionStr; 
-  /// @brief the bounds for each selection lower and upper
-  std::vector< std::vector<double> > SelectionBounds;
+  std::vector< std::string > SelectionStr;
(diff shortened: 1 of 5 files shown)


samplePDF/samplePDFFDBase.h
```diff
@@ -285,9 +285,7 @@ class samplePDFFDBase :  public samplePDFBase
   /// this is of length 3: 0th index is the value, 1st is lower bound, 2nd is upper bound
   std::vector< std::vector<double> > StoredSelection;
   /// @brief the strings grabbed from the sample config specifying the selections
-  std::vector< std::string > SelectionStr; 
-  /// @brief the bounds for each selection lower and upper
-  std::vector< std::vector<double> > SelectionBounds;
```
EdAtkin, line 290: SelectionBounds just wasn't used anywhere I guess... :face_with_diagonal_mouth:

EdAtkin (approved):
Yep this all looks really nice!

PR #414, opened 2025-04-25
feat: Minor RAM tweaks for spline

KSkwarczynski:
# Pull request description

## Changes or fixes
- There were no clean vector for 7D thus index wasn't properly delete
- UniqueSystIndices wasn't dleted at all
- TSpline3_red should delte my spline so added protection, we were dleting nullptr, I don't have a clue why code was fine...
- Trying to pass strings and vectors via reference


EdAtkin (approved):
Thanks :) this looks like it should mop up some useless RAM

PR #417, opened 2025-04-28
bugfix: Improve Thread Sanity

KSkwarczynski:
# Pull request description
Previously there was no way to run wiht more 16 threads other than commenting out line.
Simplest way to still have error (which I think is usefull for less expirenced users) but give ability to play with for more expireinced is to make check based on OMP_NUM_THREADS enviorment


EdAtkin (approved):
Thanks, this seems sensible :)

PR #418, opened 2025-04-28
feat: Allow to override config with another config

KSkwarczynski:
# Pull request description
This allow to override default config with another config.

## Examples
example

`./bin/MCMCTutorial TutorialConfigs/FitterConfig.yaml --override TutorialConfigs/Override.yaml
`

---

### DIFF
--- a/mcmc/MaCh3Factory.cpp
+++ b/mcmc/MaCh3Factory.cpp
@@ -36,6 +36,26 @@ std::unique_ptr<manager> MaCh3ManagerFactory(int argc, char **argv) {
     throw MaCh3Exception(__FILE__, __LINE__);
   }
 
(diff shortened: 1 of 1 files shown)


mcmc/MaCh3Factory.cpp
EdAtkin, line 1: Very minor but can you add a little bit more to the logger statements? e.g. the "Merging two configs" could you just say something more like "Options in override config {} will be taken over options in {}"?

Similarly for the error, can you just add something clear like "If using override feature you cannot have any additional arguments".

Sorry very nit picky.

KSkwarczynski: Sure, better error comments can't hurt 👍

PR #425, opened 2025-04-30
feat: Shared NuOscillator object for fixed binning across all samples

dbarrow257:
# Pull request description
If all samples use the same evaluation points for oscillation probability, instead of repeating the calculation nTimes, just use a shared NuOscillator object. This does not affect previous behaviour, only addition.

## Examples

---

### DIFF
--- a/samplePDF/samplePDFFDBase.cpp
+++ b/samplePDF/samplePDFFDBase.cpp
@@ -1,5 +1,7 @@
 #include "samplePDFFDBase.h"
+#include "manager/MaCh3Logger.h"
 #include "samplePDF/Structs.h"
+#include <cstddef>
 
 _MaCh3_Safe_Include_Start_ //{
 #include "Oscillator/OscillatorFactory.h"
@@ -36,8 +38,12 @@ samplePDFFDBase::samplePDFFDBase(std::string ConfigFileName, covarianceXsec* xse
 
   samplePDFFD_array = nullptr;
   samplePDFFD_data = nullptr;
-  SampleDetID = "";
+  SampleName = "";
   SampleManager = std::unique_ptr<manager>(new manager(ConfigFileName.c_str()));
+
+  // Variables related to MC stat
+  FirstTimeW2 = true;
+  UpdateW2 = false;
 }
 
 samplePDFFDBase::~samplePDFFDBase()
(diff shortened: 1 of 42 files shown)


samplePDF/samplePDFFDBase.cpp
```diff
@@ -59,7 +64,7 @@ samplePDFFDBase::~samplePDFFDBase()
   if(samplePDFFD_data != nullptr){delete[] samplePDFFD_data;}
  
   for (unsigned int iCalc=0;iCalc<NuOscProbCalcers.size();iCalc++) {
-    delete NuOscProbCalcers[iCalc];
+    if (NuOscProbCalcers[iCalc] != nullptr) delete NuOscProbCalcers[iCalc];
```
KSkwarczynski, line 67: Can you set NuOscProbCalcers to nullptr after delete?
I worry that one sample PDF delete and another will try delete as well casusing segfault.

samplePDF/samplePDFFDBase.cpp
```diff
@@ -203,9 +209,18 @@ void samplePDFFDBase::Initialise() {
...
+    MACH3LOG_INFO("Setting up NuOscillator.. ");
+    if (NuOscProbCalcers.size() != 0) {
+      MACH3LOG_INFO("You have passed an OscillatorBase object through the constructor of a samplePDFFDBase object - this will be used for all oscillation channels");
+      MACH3LOG_INFO("Overwriting EqualBinningPerOscChannel = true");
+      EqualBinningPerOscChannel = true;
+    } else {
+      InitialiseNuOscillatorObjects();
+    }
+    SetupNuOscillatorPointers();
+    OscParams = OscCov->GetOscParsFromSampleName(SampleName);
```
KSkwarczynski, line 221: I don't have strong opinion but moving all of this into InitialiseNuOscillatorObjects() would make more sense.

Conceptually all NuOsc things being set InitialiseNuOscillatorObjects make a lot of sense to me

KSkwarczynski (approved):
I love this :)

I wonder if it need some protection. For example NuOsc checks that passed pointer also is using subsampling/binned as config would suggest. But I don't think this should posptone merging

samplePDF/samplePDFFDBase.cpp
dbarrow257: So the InitialiseNuOscillatorObjects function in samplePDF initialises individual samples NuOscillator objects (which we don't want if we are using a shared object). That was my logic.

The rest of the pointer setting etc. is independent of how the NuOscillator object is initialised.

PR #427, opened 2025-05-01
bugfix: Fix Shared NuOsc destructor

KSkwarczynski:
# Pull request description
I was testing SharedNuOsc destructor here: https://github.com/mach3-software/MaCh3Tutorial/pull/117

Long term goal I think we should use shared_pointer for such cases but do this small hack for now to avoid segfault in desctuctor.

## Examples
image

---


dbarrow257 (commented):
In this situation, where do you actually delete the shared instance of the NuOscillator object?

KSkwarczynski:
I don't have much expirence with Shared Pointers but if I understood they delete after last instance of shared pointer is deleted.

So if we do Shared NuOsc, it should be deleted automatically at the end of exe.
If you initalsie in SamplePDF then it should dleete in destructor.

I think, I haven't tested it that's why it is more for future.

dbarrow257 (approved):
LGTM

PR #429, opened 2025-05-02
breaking: Fix Setting initial steps for adaption when starting from previous chain

KSkwarczynski:
# Pull request description
When you start from previous you should also set numebr of steps for adaption matrices. I don't think this affected anyone but should be fixed asap

## Changes or fixes
- Removed `setInitialStepNumber` people should use StartFromPrevious chain fucnitonality

## Examples

---


EdAtkin (approved):
thanks for catching this!

PR #430, opened 2025-05-02
feature:Tiny efficiency improvement

henry-wallace-phys:
# Pull request description
Tiny efficiency improvement, now skips the internal likelihood calc loop if parameter is flat


KSkwarczynski (approved):
Thank you Dr Wallace

PR #440, opened 2025-05-10
feat: ROOT Dicitonary to recude copy pasting

KSkwarczynski:
# Pull request description
We have feature which allow to dump Spline Mono into ROOT file. It was bit copy pasted...
Now use ROOT dictionary which allow to save class and load it much simpler. If we would like to change class in the future then we will not have to modify actual laoding bit :)

I will expand usage of it to other stucts we have.

## Examples
image
image
image
image

---


splines/CMakeLists.txt
EdAtkin, line 1: did you forget to add the MaCh3SplineDict.cxx file?

EdAtkin: nvm, there's a macro that generates it I see

EdAtkin (approved):
this looks really great and I agree we should think about if we can expand this to other structs or objects to make any loading quicker.

PR #443, opened 2025-05-13
tidy: Increase usage of groups in Doxygen 

KSkwarczynski:
# Pull request description
We were using defgroup for some time but very restrictilvey. Using renaming as oprtunity to refurbish this featue.
image

Previously we had setters which was confusging. I decided to have ParameterHandlerSetters and sperate for SampleHandler.
This should make navigation easier. We could expand to not only have setters and getters but also interface fucntion, expand to other classes etc.

## Examples

---


EdAtkin (approved):
LGTM

PR #447, opened 2025-05-13
tidy: Avoid code duplicates in Sample Handler

KSkwarczynski:
# Pull request description

## Changes or fixes
- Make new fucntion for inilaising arrays to avoid doing same in several places.
- When chencking kiematic bounds do not call ReturnKinematicParameter twice [I know it is setup], it is slower but I also find code hard to grasp
- Slilghtly hcange logic of fill_array. Stop calling _MP from normal FillArray. Now simply by looking at Reweight peopel should have better understanding what is going on

## Examples

---

### DIFF
--- a/Samples/SampleHandlerFD.h
+++ b/Samples/SampleHandlerFD.h
@@ -171,7 +171,7 @@ public:
   /// @brief wrapper to call set binning functions based on sample config info
   void SetupSampleBinning();
   /// @brief Initialise data, MC and W2 histograms
-  void SetupPDF(const size_t numberXBins, const size_t numberYBins);
+  void SetupReweightArrays(const size_t numberXBins, const size_t numberYBins);
 
   /// @brief the strings associated with the variables used for the binning e.g. "RecoNeutrinoEnergy"
   std::string XVarStr, YVarStr;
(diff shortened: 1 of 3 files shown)


Samples/SampleHandlerFD.h
```diff
@@ -170,6 +170,9 @@ class SampleHandlerFD :  public SampleHandlerBase
   void Set2DBinning(std::vector<double> &XVec, std::vector<double> &YVec) {Set2DBinning(XVec.size()-1, XVec.data(), YVec.size()-1, YVec.data());};
   /// @brief wrapper to call set binning functions based on sample config info
   void SetupSampleBinning();
+  /// @brief Initialise data, MC and W2 histograms
+  void SetupPDF(const size_t numberXBins, const size_t numberYBins);
```
EdAtkin, line 174: Do we want to call this "SetupPDF"? Given that we've just removed PDF from everywhere in the code base? I would probably call it something like "SetupReweightArrays()" or something?

KSkwarczynski: Agreed and done, to be fair I didn't think a lot when naming it

PR #448, opened 2025-05-13
tidy: Adaptive Handler into pointer

KSkwarczynski:
# Pull request description
Having adaptive MCMC as pointer means several things
- No longer allocate memory if we don't use adaptive
- Potetnially faster code due to caches
- Most importantly, in future we might want to have proper adaption factory, different algorigh etc. Then pointer make it nicer.

## Examples

---


henry-wallace-phys (approved):
Nice!

dbarrow257 (approved):
LGTM

PR #449, opened 2025-05-13
breaking: Reduce Unity type weirdness

KSkwarczynski:
# Pull request description
We have Unity which was always double, but some variabels are float or double. Pointer to float cant point to doable vice versta.

Therafore to remove lot of ugly ifdefs in the code let's intorudce 
`Unity_F `-> For floats
`Unity_D `for doubel
`Unity `for flaot or double depdning on cmake flag

Marked as breaking, as this in reality changes Old Unity into Unity_D. So expeirmental specyfic code might need to adjust sliglhy.
## Examples

---


EdAtkin (approved):
LGTM

PR #450, opened 2025-05-13
feat: Oscillation Handler

KSkwarczynski:
# Pull request description
First Merge: 
https://github.com/mach3-software/MaCh3/pull/443
https://github.com/mach3-software/MaCh3/pull/443
https://github.com/mach3-software/MaCh3/pull/449

This PR creates OscillationHandler, which is simle wrapper around NuOscillator.
Similarly how we have spline handler etc.

Most of heavy lifting is still being done by NuOscillator but having seprate class allow to nicely detach parts of interface from SampelPDF.

To make it work one must merge: https://github.com/mach3-software/MaCh3Tutorial/pull/123

## Examples

---

### DIFF
--- a/Samples/OscillationHandler.cpp
+++ b/Samples/OscillationHandler.cpp
@@ -1,0 +1,15 @@
+#include "OscillationHandler.h"
+
+_MaCh3_Safe_Include_Start_ //{
+#include "Oscillator/OscillatorFactory.h"
+#include "Constants/OscillatorConstants.h"
+_MaCh3_Safe_Include_End_ //}
+
+// ************************************************
+OscillationHandler::OscillationHandler() {
+// ************************************************
+
+
+}
+
+
--- a/Samples/OscillationHandler.cpp
+++ b/Samples/OscillationHandler.cpp
@@ -6,10 +6,95 @@ _MaCh3_Safe_Include_Start_ //{
 _MaCh3_Safe_Include_End_ //}
 
 // ************************************************
-OscillationHandler::OscillationHandler() {
+OscillationHandler::OscillationHandler(const std::string& NuOscillatorConfigFile, bool BinningPerOscChannel_,
+                                       std::vector<const double*> OscParams_, const int SubChannels) {
 // ************************************************
+  EqualBinningPerOscChannel = BinningPerOscChannel_;
+  OscParams = OscParams_;
+  auto OscillFactory = std::make_unique<OscillatorFactory>();
 
+  //DB's explanation of EqualBinningPerOscChannel:
+  //In the situation where we are applying binning oscillation probabilities to a SampleHandler object, it maybe the case that there is identical binning per oscillation channel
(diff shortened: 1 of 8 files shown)


Samples/OscillationHandler.cpp
```diff
@@ -0,0 +1,100 @@
...
+  } else {
+    for (size_t iChannel = 0; iChannel < NuOscProbCalcers.size(); iChannel++) {
+      NuOscProbCalcers[iChannel]->CalculateProbabilities(OscVec);
+    }
+  }
+}
+
+
+// ************************************************
+const M3::float_t* OscillationHandler::GetNuOscillatorPointers(int Channel, int InitFlav, int FinalFlav, FLOAT_T TrueEnu, FLOAT_T TrueCosZenith) {
```
dbarrow257, line 77: I think it's fine for now, but it seems redundant to have both channel, initflav and finalflav all be passed through here. We should think of a way to condense this info down

Samples/SampleHandlerFD.cpp
```diff
@@ -1307,77 +1272,54 @@ void SampleHandlerFD::AddData(TH2D* Data) {
...
-      throw MaCh3Exception(__FILE__, __LINE__);
+    if (YAML::LoadFile(NuOscillatorConfigFile)["General"]["CalculationType"].as<std::string>() == "Unbinned") {
+      MACH3LOG_WARN("Tried using EqualBinningPerOscChannel while using Unbinned oscillation calculation, changing EqualBinningPerOscChannel to false");
+      EqualBinningPerOscChannel = false;
     }
-    NuOscProbCalcers[0]->Setup();
-  } else {
-    NuOscProbCalcers = std::vector<OscillatorBase*>(int(MCSamples.size()));
+  }
+  std::vector<const double*> OscParams = OscParHandler->GetOscParsFromSampleName(SampleName);
```
dbarrow257, line 1290: Doesn't OscParams go out of scope once the function ends or am I missing something with how vectors are just pointer wrappers?

dbarrow257 (commented):
I forget how passing vectors through functions work but doesn't OscParams go out of scope? (Too late on a Weds evening...)

Samples/OscillationHandler.cpp
KSkwarczynski: I agree, during the process of remoing OscChannel I plan to make Channel Enums and some Channel struct so I plan to come back to this

Samples/SampleHandlerFD.cpp
KSkwarczynski: Here we make copy of this vector
https://github.com/mach3-software/MaCh3/blob/1edcd4cfef7268fe494ac734c440eb9ce450b17f/Samples/OscillationHandler.cpp#L13
so it should be fine, espcieally as pointers point to values int CovOsc. Assuming we don't delete CovOsc it should be fine ;)


Also locally CI passed so I think it is safe

PR #454, opened 2025-05-14
feat: Expand CodeQL to also perform GPU compilation

KSkwarczynski:
# Pull request description
Previously we were testing with CodeQL CPU implementation. It might be wroth to also test GPU. Why not more bots

## Examples

---


dbarrow257 (approved):
LGTM

PR #455, opened 2025-05-15
breaking: Osc Channel Info

KSkwarczynski:
# Pull request description
We should get rid of OscChannel Loop structure. This PR introduces OscChannel info, this is supposed to give idea how we should retrieve info about osc channel post OscChannel loop removal.

This retreive ability to both retrieve OscChannel index as well as pointer to it

Merge this before: https://github.com/mach3-software/MaCh3Tutorial/pull/124
## Examples

---

### DIFF
--- a/Samples/FarDetectorCoreInfoStruct.h
+++ b/Samples/FarDetectorCoreInfoStruct.h
@@ -4,7 +4,7 @@
 
 #include "Samples/Structs.h"
 
-/// @brief Store info about used osc channels
+/// @brief KS: Store info about used osc channels
 struct OscChannelInfo {
   /// Name of osc channel
   std::string flavourName;
@@ -15,9 +15,12 @@ struct OscChannelInfo {
   int InitPDG;
   /// PDG of oscillated/final flavour
   int FinalPDG;
+
+  /// In case experiment specific would like to have pointer to channel after using @GetOscChannel, they can set pointer to this
+  double ChannelIndex;
 };
 
-/// @brief Get Osc Channel Index based on initial and final PDG codes
+/// @brief KS: Get Osc Channel Index based on initial and final PDG codes
 /// @param OscChannel The vector of available oscillation channels
 /// @param InitFlav Initial flavour PDG code
 /// @param FinalFlav Final flavour PDG code
@@ -44,7 +47,6 @@ struct FarDetectorCoreInfo {
   ~FarDetectorCoreInfo(){if(isNC != nullptr) delete [] isNC;}
 
   int nEvents; ///< how many MC events are there
-  double ChannelIndex;
 
   std::vector<int*> Target; ///< target the interaction was on
   std::vector<const int*> nupdg;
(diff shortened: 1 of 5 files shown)


Samples/FarDetectorCoreInfoStruct.h
```diff
@@ -1,8 +1,41 @@
...
+
+/// @brief KS: Get Osc Channel Index based on initial and final PDG codes
+/// @param OscChannel The vector of available oscillation channels
+/// @param InitFlav Initial flavour PDG code
+/// @param FinalFlav Final flavour PDG code
+/// @return Index in OscChannel vector
+inline int GetOscChannel(const std::vector<OscChannelInfo>& OscChannel, const int InitFlav, const int FinalFlav) {
+  for (size_t i = 0; i < OscChannel.size(); ++i) {
+    if (InitFlav == OscChannel[i].InitPDG && FinalFlav == OscChannel[i].FinalPDG) {
+      return static_cast<int>(i);
```
dbarrow257, line 31: Shouldn't this be:
return static_cast<int>(OscChannel[i].ChannelIndex);

if we assume that the ordering in the vector doesn't necessarily match up with the channel indexing?

Samples/SampleHandlerFD.cpp
```diff
@@ -124,14 +124,20 @@ void SampleHandlerFD::ReadSampleConfig()
   
   nSamples = static_cast<M3::int_t>(SampleManager->raw()["SubSamples"].size());
   MCSamples.resize(nSamples);
-  
+  OscChannels.reserve(nSamples);
+
   for (auto const &osc_channel : SampleManager->raw()["SubSamples"]) {
-    oscchan_flavnames.push_back(osc_channel["Name"].as<std::string>());
-    oscchan_flavnames_Latex.push_back(osc_channel["LatexName"].as<std::string>());
+    OscChannelInfo info;
```
dbarrow257, line 130: info isn't a great variable name

dbarrow257 (changes requested):
Couple of minor changes could be done

Samples/FarDetectorCoreInfoStruct.h
KSkwarczynski: Good call, right now both should be identical, but your suggestion is more robust

Samples/SampleHandlerFD.cpp
KSkwarczynski: That's fair

Samples/FarDetectorCoreInfoStruct.h
```diff
@@ -1,8 +1,41 @@
...
+  /// Fancy channel name (e.g., LaTeX formatted)
+  std::string flavourName_Latex;
+
+  /// PDG of initial flavour
+  int InitPDG;
+  /// PDG of oscillated/final flavour
+  int FinalPDG;
+
+  /// In case experiment specific would like to have pointer to channel after using @GetOscChannel, they can set pointer to this
+  double ChannelIndex;
```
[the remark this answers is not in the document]
dbarrow257: Ah I see - good point

dbarrow257 (approved):
LGTM - Thanks for the changes

PR #458, opened 2025-05-15
breaking: Remove Oscillation Channel Loop Structure

KSkwarczynski:
# Pull request description
Currently used loop strucutre is genrally bad for perofmance, as we can't have loop over all events but need seprate loop for each channel.

In addition this change should enable having mutiple samples in one object much easier.

First merge tutorial: https://github.com/mach3-software/MaCh3Tutorial/pull/125

I did run locally all CI and it worked (LLH scnas for example).
Sample PDF requeird minor tweaks due to not having now Osc channed index in output. However weights are identical

## Examples
Here are disributons made with tutrorial.
[Test_KinemPlot.pdf](https://github.com/user-attachments/files/20229728/Test_KinemPlot.pdf)

Weights are identical, code finds issue becasue I removed osc channel from output...
example

---

### DIFF
--- a/Samples/SampleHandlerFD.cpp
+++ b/Samples/SampleHandlerFD.cpp
@@ -183,15 +183,13 @@ void SampleHandlerFD::Initialise() {
   //Now initialise all the variables you will need
   Init();
 
+  nEvents = SetupExperimentMC();
 
-  int TotalMCEvents = SetupExperimentMC();
-  MCSamples.nEvents = TotalMCEvents;
-
-  InitialiseSingleFDMCObject(MCSamples.nEvents);
+  InitialiseSingleFDMCObject(GetNEvents());
   SetupFDMC();
 
   MACH3LOG_INFO("=============================================");
-  MACH3LOG_INFO("Total number of events is: {}", TotalMCEvents);
+  MACH3LOG_INFO("Total number of events is: {}", GetNEvents());
 
   if (OscParHandler) {
     MACH3LOG_INFO("Setting up NuOscillator..");
@@ -381,7 +379,7 @@ void SampleHandlerFD::FillArray() {
   
   PrepFunctionalParameters();
 
-  for (int iEvent=0;iEvent<MCSamples.nEvents;iEvent++) {
+  for (unsigned int iEvent=0;iEvent<GetNEvents();iEvent++) {
     ApplyShifts(iEvent);
 
     if (!IsEventSelected(iEvent)) {
@@ -509,7 +507,7 @@ void SampleHandlerFD::FillArray_MP()  {
     // We will hit <0.1 s/step eventually! :D
     
     #pragma omp for
-    for (int iEvent=0;iEvent<MCSamples.nEvents;iEvent++) {
+    for (unsigned int iEvent=0;iEvent<GetNEvents();iEvent++) {
       //ETA - generic functions to apply shifts to kinematic variables
       // Apply this before IsEventSelected is called.
       ApplyShifts(iEvent);
@@ -672,7 +670,7 @@ void SampleHandlerFD::SetupFunctionalParameters() {
   // which calls RegisterIndividualFuncPar to populate funcParsNamesMap, funcParsNamesVec, and funcParsFuncMap
   RegisterFunctionalParameters();
   funcParsMap.resize(funcParsNamesMap.size());
-  funcParsGrid.resize(MCSamples.nEvents);
+  funcParsGrid.resize(GetNEvents());
 
   // For every functional parameter in XsecCov that matches the name in funcParsNames, add it to the map
   for (FunctionalParameter & fp : funcParsVec) {
@@ -693,7 +691,7 @@ void SampleHandlerFD::SetupFunctionalParameters() {
 
   // Mostly the same as CalcXsecNormsBins
   // For each event, make a vector of pointers to the functional parameters
-  for (std::size_t iEvent = 0; iEvent < static_cast<std::size_t>(MCSamples.nEvents); ++iEvent) {
+  for (std::size_t iEvent = 0; iEvent < static_cast<std::size_t>(GetNEvents()); ++iEvent) {
     // Now loop over the functional parameters and get a vector of enums corresponding to the functional parameters
     for (std::vector<FunctionalParameter>::iterator it = funcParsVec.begin(); it != funcParsVec.end(); ++it) {
       // Check whether the interaction modes match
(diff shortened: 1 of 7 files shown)


Samples/SampleHandlerFD.cpp
```diff
@@ -702,93 +693,90 @@ void SampleHandlerFD::SetupFunctionalParameters() {
...
+  for (int iSpline = 0; iSpline < MCSamples.nxsec_spline_pointers[iEvent]; ++iSpline) {
+    spline_weight *= *(MCSamples.xsec_spline_pointers[iEvent][iSpline]);
   }
   return spline_weight;
 }
 
 // ***************************************************************************
-// Calculate the normalisation weight for one event
-M3::float_t SampleHandlerFD::CalcWeightNorm(const int iSample, const int iEvent) const {
+// Calculate the normalisation weight for on event
```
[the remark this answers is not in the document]
KSkwarczynski: not a clue what happened

Samples/SampleHandlerFD.cpp
```diff
@@ -1297,70 +1271,73 @@ void SampleHandlerFD::InitialiseNuOscillatorObjects() {
   Oscillator = std::make_shared<OscillationHandler>(NuOscillatorConfigFile, EqualBinningPerOscChannel, OscParams, static_cast<int>(OscChannels.size()));
 
   if (!EqualBinningPerOscChannel) {
-    for (size_t iSample=0;iSample<MCSamples.size();iSample++) {
+    for(int iChannel = 0; iChannel < static_cast<int>(OscChannels.size()); iChannel++) {
       std::vector<M3::float_t> EnergyArray;
       std::vector<M3::float_t> CosineZArray;
 
-      for (int iEvent=0;iEvent<MCSamples[iSample].nEvents;iEvent++) {
+      for (int iEvent = 0; iEvent< MCSamples.nEvents;iEvent++) {
```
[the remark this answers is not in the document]
KSkwarczynski: You are right, haven't benchmarked but this is setup so probably fine.
Even if it's 6 times slower I don't think this operation is slow to be actual bottleneck for setup [In my expirence spline loading is the slowest]

KSkwarczynski:
> One quick thought, is it ever helpful to know the event index within an oscillation channel and should we add the functionality to retain this? e.g. event 678,942 caused a problem, you could then have to go figure out what happened to this event but you don't have the relative index for that event in say a fhc_numu_x_numu.root file?

So I haven't added Channel index, Dan made a point that we need Initial and FinalFalvour so there is no need for third int.
If this is error you could simplgy call `GetOscChannel`

There is also ability to get the pointer to value of osc channel
https://github.com/mach3-software/MaCh3/blob/c22c68bd253b98c99c6a5d8301bc79fbcdf65243/Samples/SampleHandlerFD.cpp#L1946C1-L1952C2

Samples/SampleHandlerFD.cpp
```diff
@@ -810,41 +798,33 @@ void SampleHandlerFD::SetupNormParameters() {
...
   #endif
 }
 
 // ************************************************
 //A way to check whether a normalisation parameter applies to an event or not
-void SampleHandlerFD::CalcNormsBins(const int iSample) {
+void SampleHandlerFD::CalcNormsBins() {
 // ************************************************
-  FarDetectorCoreInfo *fdobj = &MCSamples[iSample];
+  FarDetectorCoreInfo *fdobj = &MCSamples;
```
dbarrow257, line 827: This now seems somewhat unneeded

Samples/SampleHandlerFD.cpp
```diff
@@ -1120,57 +1096,54 @@ void SampleHandlerFD::Set2DBinning(size_t nbins1, double low1, double high1, siz
...
-      MCSamples[mc_i].rw_upper_upper_xbinedge[event_i] = upper_upper_edge;
+  for(int event_i = 0 ; event_i < MCSamples.nEvents ; event_i++){
+    //Set x_var and y_var values based on XVarStr and YVarStr
+    MCSamples.x_var[event_i] = GetPointerToKinematicParameter(XVarStr, event_i);
+    MCSamples.y_var[event_i] = GetPointerToKinematicParameter(YVarStr, event_i);
+
+    //Global bin number
+    int bin = _hPDF2D->FindBin(*(MCSamples.x_var[event_i]), *(MCSamples.y_var[event_i]));
+
+    int bin_x = -999;
```
dbarrow257, line 1107: M3::__BAD_INT__?

Samples/SampleHandlerFD.cpp
```diff
@@ -183,18 +183,13 @@ void SampleHandlerFD::Initialise() {
...
-    MCSamples[iChannel].nEvents = SetupExperimentMC(iChannel);
-    MACH3LOG_INFO("Number of events processed: {}", MCSamples[iChannel].nEvents);
-    TotalMCEvents += MCSamples[iChannel].nEvents;
-    MACH3LOG_INFO("Initialising FDMC object..");
-    InitialiseSingleFDMCObject(iChannel, MCSamples[iChannel].nEvents);
-    SetupFDMC(iChannel);
-    MACH3LOG_INFO("Initialised channel: {}/{}", iChannel, OscChannels.size());
-  }
+
+  int TotalMCEvents = SetupExperimentMC();
```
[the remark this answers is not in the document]
dbarrow257: Would also allow us to use it instead of MCSamples.nEvents

dbarrow257 (commented):
Minor comments relating to changing flow of variables through code

Samples/SampleHandlerFD.cpp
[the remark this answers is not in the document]
KSkwarczynski: Now nEvents is in samplePDFBase, GetNEvents is no longer virtual. tl:dr done

Samples/SampleHandlerFD.cpp
KSkwarczynski: This was done to redcue number of changes in PR to make review easier.
I can tidy this but in next PR for transparency

Samples/SampleHandlerFD.cpp
KSkwarczynski: done

Samples/SampleHandlerFD.h
```diff
@@ -357,7 +345,7 @@ class SampleHandlerFD :  public SampleHandlerBase
   std::unique_ptr<manager> SampleManager;
   /// @brief function to create the member of the FarDetectorInfo struct so
   /// they are the appropriate size.
-  void InitialiseSingleFDMCObject(int iSample, int nEvents);
+  void InitialiseSingleFDMCObject(const int nEvents);
```
dbarrow257, line 348: With the latest change, I think you can remove the argument from the function signature

dbarrow257 (commented):
Another comment regarding function signatures

Samples/SampleHandlerFD.h
dbarrow257: I guess you can also remove fdmc->nEvents as well

KSkwarczynski: done

dbarrow257 (approved):
LGTM now, thanks for the changes!

PR #466, opened 2025-05-19
feat: Make Sure shared NuOscillator has same number of paramters

KSkwarczynski:
# Pull request description
Within old implemntation it was techcically possible to pass ATM Oscillator to Beam. Which would casue error when running NuOscillator.

Problem is it would be harder to spot.

Also tidy nEvents

## Examples

---


dbarrow257 (approved):
LGTM - Sensible change

PR #468, opened 2025-05-20
tidy:

KSkwarczynski:
# Pull request description
Simply split Structs into SampleStructs and ParameterStructs.

Having most of Paramter structs in Sample/Structs was bit silly

## Examples

---


EdAtkin (approved):
Minor but do we want to call these files "SampleStructs" or "SampleUtils" ? These is incredibly nit picky but technically there aren't just structs in these files...

PR #472, opened 2025-05-21
tidy: Do not allow adpation and PCA

KSkwarczynski:
# Pull request description
Right now one could run adaption and PCA. Which would not work in a sense hard to predict outcome as parmaters in PCA base doon't use throw matrix, while there is no itnerface to use Adaption for undecomposed params. Adaption and PCA handler are independet.

Might be worth to fix it in future [if somone has time]. Right now simply throw error.

## Examples

---


dbarrow257 (approved):
LGTM

PR #474, opened 2025-05-22
tidy: Minor update to logger message 

yash-quark:
# Pull request description
A minor logger message update: Notify the user when you only supply YVarStr without an XVarStr in the FD sample yaml that the code will crash.

Note: This is not to be confused with checking whether there are no bins at all. 

## Examples

---


KSkwarczynski (approved):
Thanks

PR #480, opened 2025-05-27
feat: Expand usage of Param Group

KSkwarczynski:
# Pull request description
As we expand number of uncertanties we need better way to control parmates from various groups (xsec, flux and osc in future)

## Changes or fixes
- Now we can easily bet parmater index using name
- SetGroupOnlyParameters also support setting values from vecto
- SetGroupOnlyParameters also can pass mutlple groups

## Examples

---


EdAtkin (approved):
Again, looks nice :ok_hand:

PR #482, opened 2025-05-28
feat: CI build to work with forks

KSkwarczynski:
# Pull request description
MaCh3 CI was built to work only for people commiting direclty to MaCh3.
Many open projects (Like Nuisance, ROOT) allow people to make cotnribtion from forks.

This is attempt for at least part of CI to work with forks. 

## Examples

---


pjdunne:
Copying comment from slack here. I think slack is a better place for the discussion of this one as it's not about the code so let's continue there:

"Hi 
[@Kamil Skwarczyński], we’ve generally discouraged people from working with forks in the past as it’s led to eg the original atmospheric fit fork that was described as using ‘mach3’ for some time without taking any of our updates.

While it’s good to try and be more inclusive I think making it easy to use forks would be a change to our recommendations on how to contribute that we should discuss.

Could we hold off from merging this until that discussion concludes please."

PR #484, opened 2025-05-29
Update leftover GetName to GetTitle in data histogram handling

yash-quark:
# Pull request description
This is just a quick and minor addition to [#481](https://github.com/mach3-software/MaCh3/pull/481). In the former, I only changed GetName to GetTitle in the logger of SampleHandlerFD and not in the Base class which also uses the same.

## Examples

---


KSkwarczynski (approved):
LGTM

PR #497, opened 2025-06-05
tidy: Avoid doing ==xsec_cov in FitterBase

KSkwarczynski:
# Pull request description
In the past only xsec_cov had fancy name. Now as we unify more and more Covariance Objects this is no longer true.

And "name ==xsec_cov" is very cofnusing for new users.

## Examples

---

### DIFF
--- a/Fitters/FitterBase.cpp
+++ b/Fitters/FitterBase.cpp
@@ -1,4 +1,5 @@
 #include "FitterBase.h"
+#include "Parameters/ParameterHandlerGeneric.h"
 
 _MaCh3_Safe_Include_Start_ //{
 #include "TRandom.h"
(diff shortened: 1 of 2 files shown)


dbarrow257:
It looks like the only use of the isXsec check was for retrieving the FancyName?

Thinking about it, wouldn't the covBase object benefit from having that ability?

dbarrow257:
> So CovBase has fancy name.
> 
> https://github.com/mach3-software/MaCh3/blob/daa1e55ee09eae7ef6f29e11911d9817dd6cec85/Parameters/ParameterHandlerBase.h#L171
> 
> My reasoning was that for T2K we have plenty matrices which do not have fancy name filled: like this https://github.com/t2k-software/MaCh3/blob/develop/covariance/covarianceCorrelatedFD.cpp and https://github.com/t2k-software/MaCh3/blob/develop/covariance/covarianceSKDetBeam.cpp
> 
> But maybe it is better to do as you said and update T2K covariances?

In which case, yeah I think we just require a fancy name to be defined.
I guess a "simple" approach could be just to copy the name into fancyName if a fancyName is not defined?

dbarrow257 (approved):
LGTM - Thanks!

PR #502, opened 2025-06-06
feat: Flexible Config getter from chain

KSkwarczynski:
# Pull request description
Old ways of having multiple matrices requeir us to hack matcis name to be xsec_cov etc. Since Core assumed such naming cnvention it forced experiemnt specyfic to obey.

This PR introduces new getter so if there is only one TMacro [coming from GenericHandler] only this will be taken. Otherwise code will use old school hardcdoing for backward compatibility.

## Examples
test

[Test_drawCorr.pdf](https://github.com/user-attachments/files/20625881/Test_drawCorr.pdf)

image

[Test.root_PenaltyTerm.pdf](https://github.com/user-attachments/files/20626433/Test.root_PenaltyTerm.pdf)

---

### DIFF
--- a/Diagnostics/GetPenaltyTerm.cpp
+++ b/Diagnostics/GetPenaltyTerm.cpp
@@ -1,6 +1,7 @@
 // MaCh3 includes
 #include "Manager/Manager.h"
 #include "Samples/SampleStructs.h"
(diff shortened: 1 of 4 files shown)


dbarrow257 (approved):
LGTM

PR #508, opened 2025-06-09
feat: Store More info In Chain output

KSkwarczynski:
# Pull request description
Now we also store Data hisotrgm and Config file for each stored sample

## Examples

test

---


dbarrow257 (commented):
LGTM -  Although I'm probably missing something.

In CombineMaCh3Chains, do we compare the data histograms between different chains to ensure they are identical?

PR #509, opened 2025-06-10
feat: Port Jarlskog

KSkwarczynski:
# Pull request description
This port Jarlskog from T2K repo. Original code was quite hacky so here I rmevoed quite a lot of hacks.

Two fucnionalies first just add Jarlskog to MCMCPorcessor root file. Second used output for plotting and boht adds 

Test were done using Tutorila which do not have "T2K-like" shape...

## Examples
image
image
image

---

### DIFF
--- a/Diagnostics/ProcessMCMC.cpp
+++ b/Diagnostics/ProcessMCMC.cpp
@@ -1,21 +1,22 @@
 //MaCh3 includes
-#include "Fitters/MCMCProcessor.h"
+#include "Fitters/OscProcessor.h"
 #include "Manager/Manager.h"
 
 /// @file ProcessMCMC.cpp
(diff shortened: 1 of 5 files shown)


dbarrow257 (commented):
Maybe I'm being really dense (it's been a long week...)

Why do we need to calculate the Jarlskog when we already have the "Chain->SetAlias("J_cp",..."?

dbarrow257:
If you can put a github issue that we need to move to one approach somewhere on the tracker, I can approve these changes as they look sensible.

KSkwarczynski:
Ok added here:
image
https://github.com/orgs/mach3-software/projects/1/views/1

dbarrow257 (approved):
LGTM - Thanks.

Absolutely not for this PR, but we should try and think of a way to include support for more complex BSM related variables (looking at the Sin2Theta13Index for example).

PR #511, opened 2025-06-11
breaking: Return Projection binning from yaml

KSkwarczynski:
# Pull request description
This PR modifes how return binning for prohection works. No longer it is hardcoded in experiemnt specyfic but now it is handled via yaml.

You can see comment here:
https://t2k-experiment.slack.com/archives/C08MMSCFWDV/p1749571829614129

This has been tested on tutorial here: https://github.com/mach3-software/MaCh3Tutorial/pull/140
## Examples

[Test_KinemPlot.pdf](https://github.com/user-attachments/files/20690175/Test_KinemPlot.pdf)

---

### DIFF
--- a/python/samples.cpp
+++ b/python/samples.cpp
@@ -183,16 +183,6 @@ public:
         );
     }
     
-    std::vector<double> ReturnKinematicParameterBinning(std::string) override {
-        PYBIND11_OVERRIDE_PURE_NAME(
-            std::vector<double>,          /* Return type */
-            SampleHandlerFD,              /* Parent class */
-            "get_event_kinematic_binning",/* python name*/
-            ReturnKinematicParameterBinning, /* Name of function in C++ (must match Python name) */
-            py::arg("variable")           /* Argument(s) */
-        );
-    }
-
     void RegisterFunctionalParameters() override {
         PYBIND11_OVERRIDE_PURE_NAME(
             void,


Samples/SampleHandlerFD.cpp
```diff
@@ -119,10 +119,11 @@ void SampleHandlerFD::ReadSampleConfig()
     throw MaCh3Exception(__FILE__, __LINE__);
   }
   
-  //FD file info
-  if (!CheckNodeExists(SampleManager->raw(), "InputFiles", "mtupleprefix")){
-    MACH3LOG_ERROR("InputFiles:mtupleprefix not given in {}, please add this", SampleManager->GetFileName());
+  if (!CheckNodeExists(SampleManager->raw(), "BinningFile")){
```
dbarrow257, line 122: Do we think that the binning projections should be a sample-specific thing or a general yaml thing?

Things like XBin, YBin variables and OscChan and ModeIndex will be sample specific... but we already know what these are in core and can return some sensible binning before jumping into the exp-specific binning return function?

I don't really know.. maybe for things like Q2, W2, two samples might have very different distributions so should be sample specific...

This is really just noise 😅

Samples/SampleHandlerFD.cpp
```diff
@@ -1751,6 +1752,41 @@ std::string SampleHandlerFD::ReturnStringFromKinematicVector(const int Kinematic
   return "";
 }
 
+// ************************************************
+std::vector<double> SampleHandlerFD::ReturnKinematicParameterBinning(const std::string& KinematicParameter) {
```
dbarrow257, line 1756: I really don't think it's necessary for this PR - but it would be really nice to include some functionality that if the binning isn't specified for some variable, it defaults to use ROOTs auto binning: https://root.cern/doc/master/classTH1.html#auto-bin

dbarrow257 (approved):
LGTM: I've left some comments - I think the PR is fine as is, but we should have a further think about potential improvements.

Samples/SampleHandlerFD.cpp
KSkwarczynski: We kind of support both
https://github.com/mach3-software/MaCh3Tutorial/blob/5a6240954d0a91e361bdaea99e7c627d8d70f373/TutorialConfigs/Samples/TutorialBinning.yaml#L1C1-L10C75

We will support sample specyfic binning, one reason is one you outlined second is once we have mutliple samples in one object.

I also preserved option to use "genral" binning to bascially maintain current approach as it is before this PR.

PR #525, opened 2025-06-24
feat: Do not apply special proposal for fixed params

KSkwarczynski:
# Pull request description
Dan raised issue that even if we fixed param code might still apply special proposal like flipping.
Just to be completly safe turn off speical proposal for fixed params

## Examples

---


dbarrow257 (commented):
How will this work with the workflow where you fix parameters after a covXsec object is initialised?

I think (but could be very wrong) this will fail for cases like this: https://github.com/t2k-software/MaCh3/blob/44e769a5742cc20a34b8935008b63bbc1cb06879/samplePDF/MaCh3FactoryT2K.cpp#L14-L15

PR #563, opened 2025-07-15
bugfix: Stop Accessing out of bound knot

KSkwarczynski:
# Pull request description
If in loop
for (int k = 0; k < nPoints; ++k) {
you acess 
k+1
`spline->GetCoeff(k+1, x2, y2, b2, c2, d2);`
you are getting out of bounds.

Here are some debug messgaes to show that I was getting undefined knot
```
[DEBUG] GetCoeff(k=1): x1=1, y1=0.801279
[DEBUG] GetCoeff(k+1=2): x2=2, y2=0.801279
[DEBUG] Slope (tempb) = 0
[DEBUG] Stored: XPos[1] = 1, YResp[1] = 0.801279, Par[1][0] = 0
[DEBUG] GetCoeff(k=2): x1=2, y1=0.801279
[DEBUG] GetCoeff(k+1=3): x2=6.91753e-310, y2=3.13151e-294
```

Modification take same coeefficient as in previous segment.

While tesintg LLH I didn't notice difference. Which menst we are not using last segment [which is out of bound kind of..].

This means there is potential to reduce number of segments we store but haven't played with this more [this would requeir modyfing monolith]

TL:DR Everything is fine although code was silly

## Examples
example of before and after [modifed LLH scan range] no impact on LLH
image

image

---

### DIFF
--- a/.github/workflows/Greetings.yml
+++ b/.github/workflows/Greetings.yml
@@ -9,7 +9,7 @@ jobs:
       issues: write
       pull-requests: write
     steps:
-    - uses: actions/first-interaction@v1
+    - uses: actions/first-interaction@v2
       with:
         repo-token: ${{ secrets.GITHUB_TOKEN }}
         issue-message: "This is your first issue, thank you for contributing to MaCh3!"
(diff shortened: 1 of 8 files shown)


dbarrow257 (approved):
LGTM

PR #589, opened 2025-08-13
bugfix: Reset Step Size when loading adapt matrix

KSkwarczynski:
# Pull request description
After chatting with Abi let's try to reset step size when loading adapt.

We do reset step size when we enable adapt. However then when we load matrix step size is not rested i.e. adaption may not work as intended.

## Examples

---


dbarrow257 (approved):
LGTM

henry-wallace-phys (approved):
LGTM

PR #597, opened 2025-08-19
feat: Allow to decrease entires in MCMC Processor

KSkwarczynski:
# Pull request description
Brought up during discussion with Marvin. This allows to set entries and this way if your chain is huge you make it suable with restricted RAM

## Examples

---

### DIFF
--- a/Diagnostics/DiagMCMC.cpp
+++ b/Diagnostics/DiagMCMC.cpp
@@ -23,7 +23,9 @@ void DiagMCMC(const std::string& inputFile, const std::string& config)
   Processor->SetnLags(GetFromManager<int>(Settings["DiagMCMC"]["nLags"], 25000));
   Processor->SetPrintToPDF(GetFromManager<bool>(Settings["PrintToPDF"], true));
   Processor->Initialise();
-
+  if(Settings["MaxEntries"]) {
+    Processor->SetEntries(Get<int>(Settings["MaxEntries"], __FILE__, __LINE__));
+  }
   //KS: finally call main method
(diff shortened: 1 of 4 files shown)


dbarrow257 (commented):
I'm maybe missing something obvious but I think the logic might not work so well?

nEntries = number of Entries in TChain (https://github.com/mach3-software/MaCh3/blob/f92a12672668d6c6ca1efec64cbf641565b48006/Fitters/MCMCProcessor.cpp#L1955)

In the SetEntries function, you update nEntries to be equal to the argument of the function. And we loop from 0 to nEntries in the TChain (e.g. https://github.com/mach3-software/MaCh3/blob/f92a12672668d6c6ca1efec64cbf641565b48006/Fitters/MCMCProcessor.cpp#L861)

Say we wanted 2 million steps, but have a burnin of 100k. 

The logic would be loop from 0 to 2M but skip any event with step < 100k. Consequently we'd only have 1.9M events, despite asking for 2M.

Maybe this is ok providing we document the option more clearly?

Similarly, what happens for many hadded chains? Imagine you 20 hadded chains together?

Chain 0 step 0, Chain 0 step 1, .... Chain 0 step N, Chain 1 step 0, Chain 1 step 1, ... , Chain M step N. 

If N < burnin and M * N > nEntries, the logic we have would have 0 accepted steps.

Contrived example, and in reality I don't think this is really a problem, but we should make some clear documentation

dbarrow257 (approved):
LGTM - Thanks for the changes

PR #598, opened 2025-08-19
feat: Introduce Global bin number offset

KSkwarczynski:
# Pull request description
When we add multiple samples to one object we need a way to retain ability to use 1d `SampleHandlerFD_array_w2[i]`

However, to make this work we need to be able to set offset.

To understand what I mean, follow this example
```
Sample 0 : bins 10, offstet 0
Sample 1: bins 20, offset 10
Sample 2: bins 5, offstet 30
```

MaCh3 will never support bin edges which can change during run time so this is safe.
Even if even will migrate to different sample (assuming we want support it) this will work.

---


KSkwarczynski:
So I was initially setting it to be vector because if sample 0 has 2D while sample 1 3D and Sample 2 PolyStyle then one object is bit difficult to make.

But I think no matter what approach we decide, we would need GlobalBin Offset?

But I now better understand what you meant.

dbarrow257 (approved):
LGTM

PR #624, opened 2025-09-09
feat: Add ability to temporarily disable flipping in delayed rejection

henry-wallace-phys:
# Pull request description
Add ability to flip on the first delayed step and then NO OTHER STEPS for that given delay. Allows for small adjustments to try and find a mode

## Changes or fixes
-> Add interface to disable flips to parameter handler base
-> Add method to restore flips to DelayedRejection

---

### DIFF
--- a/Fitters/DelayedMR2T2.cpp
+++ b/Fitters/DelayedMR2T2.cpp
@@ -13,8 +13,11 @@ DelayedMR2T2::DelayedMR2T2(manager * const manager) : MR2T2(manager) {
     // On delayed on out of bounds steps. This saves some compute time and possibly regains efficiency
     delay_on_oob_only = GetFromManager<bool>(fitMan->raw()["General"]["MCMC"]["DelayOnlyOutBounds"], false);
 
-    // 
+    // Delay with a certain probability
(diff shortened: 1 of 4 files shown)


Fitters/DelayedMR2T2.cpp
```diff
@@ -73,6 +83,18 @@ void DelayedMR2T2::StoreCurrentStep() {
...
         start_step_scale[i] = systematics[i]->GetGlobalStepScale();
     }
+
+    // Don't need to disable
+    if(keep_flipping_on_reject){
+        return;
+    }
+
+    // We also need the flips
+    initial_flip_setting = std::vector<bool>(systematics.size());
```
KSkwarczynski, line 93: Could you avoid using 
initial_flip_setting = std::vector<bool>(systematics.size());

this creates new vector. There is NO memory leak but size is constant.

Realisataclly you could resize once in 
`void DelayedMR2T2::PrepareOutput() {`
becasue SPEEED

KSkwarczynski (commented):
Shoulnd't you add something here?
https://github.com/mach3-software/MaCh3/blob/ad5f2fd15f8e42e7e4f7ff8a8a9366a716b6d103/Parameters/ParameterHandlerBase.cpp#L575C1-L593C2

PR #636, opened 2025-09-17
bugfix: Fix buggy canvasname in ProcessMCMC

yash-quark:
# Pull request description
If my MCMC chains are `/path/to/file1.root` and `/path/to/file2.root`, current implementation of assigning `canvasname` for ProcessMCMC produces canvases named `/path/to/file_file2.pdf`

## Changes or fixes
Correctly extract the filename so that canvasname correctly names it `/path/to/file1_file2.pdf`. This is the expected behaviour although in near future it will be better to assign a proper outpathname

## Examples

---


KSkwarczynski (approved):
LGTM

PR #637, opened 2025-09-17
feat: Speedy Cholesky decomposition

KSkwarczynski:
# Pull request description
ROOT's cholesky is enough for most fits. However for HUGE matrices like 5k that T2K is using or 3k that Abi is using it can be bit slow.

This adds new choleksy which is faster because of better cache structure and OMP. Included example in PR which used for comapring both implemantions

## Examples
2kTest
3K
SpeedyCHolesky

[TestBlarb.txt](https://github.com/user-attachments/files/22391497/TestBlarb.txt)

---


yash-quark (approved):
🐎  LGTM! Another ROOT feature purge.

PR #643, opened 2025-10-04
feat: Predictive Plotting

KSkwarczynski:
# Pull request description
Initial version of plotting script for predictive analysis.

There is still a lot I need to add so this is initial version supporting only overlay of two files.

## Examples
[Overlay_Predictive.pdf](https://github.com/user-attachments/files/22695854/Overlay_Predictive.pdf)

---

### DIFF
--- a/Plotting/PredictivePlotting.cpp
+++ b/Plotting/PredictivePlotting.cpp
@@ -1,0 +1,233 @@
+//MaCh3 Includes
+#include "plottingUtils/plottingUtils.h"
+#include "plottingUtils/plottingManager.h"
+
+//this file has lots of usage of the ROOT plotting interface that only takes floats, turn this warning off for this CU for now
+#pragma GCC diagnostic ignored "-Wfloat-conversion"
+#pragma GCC diagnostic ignored "-Wconversion"
+
+/// @file PredictivePlotting.cpp
+/// @author Kamil Skwarczynski
+
+
+constexpr Color_t PosteriorColor[] = {kBlue, kRed, kGreen};
+const std::string Titles[] = {"Prior Predictive","Posterior Predictive", "Third File"};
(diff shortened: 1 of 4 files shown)


Plotting/PredictivePlotting.cpp
```diff
@@ -0,0 +1,233 @@
...
+  pad1->SetBottomMargin(0);
+
+  pad2->SetLeftMargin(canv->GetLeftMargin());
+  pad2->SetRightMargin(canv->GetRightMargin());
+  pad2->SetTopMargin(0);
+  pad2->SetBottomMargin(0.28);
+
+  for(size_t iSample = 0; iSample < SampleNames.size(); iSample++)
+  {
+    const int nFiles = static_cast<int>(InputFiles.size());
```
dbarrow257, line 75: Can this not potentially break?

Titles (hardcoded array) gets accessed by Titles[iFile] below and there doesn't seem to be any checking that nFiles < 3?

Plotting/PredictivePlotting.cpp
```diff
@@ -0,0 +1,233 @@
...
+
+//this file has lots of usage of the ROOT plotting interface that only takes floats, turn this warning off for this CU for now
+#pragma GCC diagnostic ignored "-Wfloat-conversion"
+#pragma GCC diagnostic ignored "-Wconversion"
+
+/// @file PredictivePlotting.cpp
+/// @author Kamil Skwarczynski
+
+
+constexpr Color_t PosteriorColor[] = {kBlue, kRed, kGreen};
```
dbarrow257, line 13: It would be good to have this configurable

dbarrow257 (changes requested):
Just one comment on some potential bound breaking

PR #645, opened 2025-10-06
feat: Process mcmc reweighting plots

DavidRiley12345:
# Pull request description

This adds a separate reweighting script, ReweightMCMC.cpp, which should be used to handle all types of reweights, including 1D and 2D in the same script. The reweight can be configured using yaml config files which are then stored in the output chain as TMacros, making it easy to work out what reweight was applied to a chain.

This script creates a branch called 'Weight', which can then be picked up and used by the ProcessMCMC script when plotting chains. 

Supported reweight types are currently 1D Gaussian, 1D TGraph, and 2D TGraph. The 1D Gaussian implementation calls out to the ReweightPrior in MCMCProcessor, while the TGraph 1&2D reweightings are implemented in this script. This leads to slightly different formatting of the TMacro which is saved, to keep compatibility with other functions which make use of the MCMCProcessor reweight (such as the Jarlskog plotting in ProcessMCMC). 

## Changes or fixes

Changes to OscProcessor so that the wrong weight doesn't accidentally apply during Jarlskog plotting.

## Examples

Example of a 1D Gaussian yaml config:

image

Example output of ProcessMCMC comparing two MCMC files, one with weight and one without:

image

Example output of ProcessMCMC comparing two MCMC files both with separate weights applied.

image

---

### DIFF
--- a/Fitters/OscProcessor.cpp
+++ b/Fitters/OscProcessor.cpp
@@ -171,14 +171,15 @@ void OscProcessor::PerformJarlskogAnalysis() {
       std::string reweightType = GetFromManager<std::string>(firstReweight["ReweightType"], "");
       auto paramNames = GetFromManager<std::vector<std::string>>(firstReweight["ReweightVar"], {});
       if (dimension == 1 && reweightType == "Gaussian" && paramNames.size() == 1){
-        Sin13_NewPrior = Get<std::pair<double, double>>(firstReweight["ReweightPrior"]);
+        Sin13_NewPrior = Get<std::pair<double, double>>(firstReweight["ReweightPrior"],__FILE__,__LINE__);
         DoReweight = true;
       } else {
         MACH3LOG_INFO("No valid reweighting configuration (1D Gaussian on sin2th_13 only) found for Jarlskog analysis");
       }
-  } else {
-    MACH3LOG_INFO("No reweighting configuration found for Jarlskog analysis");
-  }
+    } else {
+      MACH3LOG_INFO("No reweighting configuration found for Jarlskog analysis");
+    }
+}
 
   TempFile->Close();
   delete TempFile;


Fitters/MCMCProcessor.cpp
```diff
@@ -3079,6 +3079,9 @@ void MCMCProcessor::ReweightPrior(const std::vector<std::string>& Names,
 
   for (int i = 0; i < nEntries; ++i)
   {
+    if(i % 500000 == 0){
```
dbarrow257, line 3082: Instead of 500000, could this be the 1/10th of the number of steps in the chain (i.e. nEntries) so it can scale with different length chains?

Diagnostics/ProcessMCMC.cpp
```diff
@@ -688,6 +688,10 @@ void ReweightPrior(MCMCProcessor* Processor)
   std::vector<double> NewCentral = Prior[1].as<std::vector<double>>();
   std::vector<double> NewError = Prior[2].as<std::vector<double>>();
 
+  for (const auto& name : Names)
+  {
+    MACH3LOG_INFO("Reweighting prior for {} with new central value {} and error {}", name, NewCentral[&name - &Names[0]], NewError[&name - &Names[0]]);
+  }
   Processor->ReweightPrior(Names, NewCentral, NewError);
```
dbarrow257, line 695: I thought the idea of the additional executable was to remove the ability to reweight the prior from the ProcessMCMC script?

Fitters/OscProcessor.cpp
```diff
@@ -161,10 +161,38 @@ void OscProcessor::PerformJarlskogAnalysis() {
...
+      }
+    }
+    // Check if we have the new ReweightMCMC structure (not used for Jarlskog)
+    else if(CheckNodeExists(Settings, "ReweightMCMC")) {
+      MACH3LOG_WARN("Found ReweightMCMC structure in Reweight_Config");
+      MACH3LOG_WARN("Jarlskog analysis expects simple format:");
+      MACH3LOG_WARN("ReweightType: \"Gaussian\"");
+      MACH3LOG_WARN("Weight:");
+      MACH3LOG_WARN("  {}: [mean, sigma]", Sin2Theta13Name);
+      MACH3LOG_WARN("Skipping reweighting for Jarlskog analysis");
```
dbarrow257, line 193: Is there a technical reason why the Jarlskog can't include the fancy new format?

dbarrow257 (changes requested):
Generally looks good - couple of minor comments left

Fitters/MCMCProcessor.cpp
DavidRiley12345: Done! And updated to use the MaCh3Utils fancy progress bar

Fitters/OscProcessor.cpp
DavidRiley12345: Actually after having a look, the Jarlskog Analysis here uses the parameters from a 1D gaussian later on, and generates random values from that gaussian for Savage Dickey i think? so this wont work if the weight isnt a 1D sin2th13 reweight, I've added that check back in.

PR #656, opened 2025-10-20
breaking: Remove variables from FarDetectorCoreInfo

KSkwarczynski:
# Pull request description
FarDetectorCoreInfo had plenty of variables used only for setup for longer thread see: 
https://t2k-experiment.slack.com/archives/C08MMSCFWDV/p1759306677972969

This will work only after merging https://github.com/mach3-software/MaCh3Tutorial/pull/180

This feature will result in around 30MB being saved for every 1M of MC events.

Did simple benchmark at this PR increase setup time by around a few percent on the other single step should be faster ~5% due to cache locality.

## Changes or fixes
- Now all return kinematic etc are const. This is breaking change but especially for operation like passing string this allow to save a few cycles.

## Examples
blarb

---

### DIFF
--- a/python/samples.cpp
+++ b/python/samples.cpp
@@ -153,7 +153,7 @@ public:
         );
     }
 
-    int ReturnKinematicParameterFromString(std::string) {
+    int ReturnKinematicParameterFromString(const std::string&) const {
         PYBIND11_OVERRIDE_PURE_NAME(
             int,                     /* Return type */
             SampleHandlerFD,            /* Parent class */
@@ -163,7 +163,7 @@ public:
         );
     }
     
-    std::string ReturnStringFromKinematicParameter(int) {
+    std::string ReturnStringFromKinematicParameter(const int) const {
         PYBIND11_OVERRIDE_PURE_NAME(
             std::string,                /* Return type */
             SampleHandlerFD,            /* Parent class */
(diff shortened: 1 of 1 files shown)


Samples/SampleHandlerFD.cpp
```diff
@@ -1127,21 +1136,27 @@ void SampleHandlerFD::InitialiseNuOscillatorObjects() {
...
+            EnergyArray.push_back(M3::float_t(ReturnKinematicParameter("TrueNeutrinoEnergy", iEvent)));
+            #pragma GCC diagnostic pop
           }
         }
         std::sort(EnergyArray.begin(),EnergyArray.end());
 
         //============================================================================
         //DB Atmospheric only part, can only happen if truecz has been initialised within the experiment specific code
-        if (*(MCSamples[0].rw_truecz) != M3::_BAD_DOUBLE_) {
+        if (ReturnKinematicParameter("TrueCosineZenith", 0) != M3::_BAD_DOUBLE_) {
```
dbarrow257: Doesn't this force every experiment to have to include 'TrueCosineZenith' in the experiment defined ReturnKinematicParameter function and get it to return M3::_BAD_DOUBLE_?

For example, when T2K code doesn't find TrueCosineZenith, it throws error: https://github.com/t2k-software/MaCh3/blob/af4d963913b89a1ee6152c6923a0b1f1cfb6cafc/SamplesT2K/SampleHandlerSK.cpp#L801-L805

Which makes more sense to me than having to "know" to return the M3::BadDouble value

PR #659, opened 2025-10-22
tidy: float/double to MCMC Processor

KSkwarczynski:
# Pull request description
`ParStep `can take quite a lot of memory. By giving ability to use float we can save some RAM when using MCMC processor.

## Changes or fixes
- increase usage of smart pointer
- There are variables like ParamSums which are only used by AutoCorr code before they we class members. Now much more variables are local. Which should make keeping track what is used where much easier

## Examples

---


dbarrow257 (approved):
LGTM

PR #660, opened 2025-10-22
tidy: Minor Bin Speedup

KSkwarczynski:
# Pull request description
Was experimenting with more optimised bin finding code. Idea is the same this is rather hyper optimisation pitfall.
Main thing is to cache local variables and instead of using` if else` just use `if` we return so it doesn't matter

## Examples
[test.txt](https://github.com/user-attachments/files/23050397/test.txt)
image

---


dbarrow257 (approved):
LGTM - One idea for a further speedup

PR #787, opened 2026-02-11
feat: Replace atomic with reduction in reweighting loop

KSkwarczynski:
# Pull request description
We were manually allocating private array and then merging them using `atomic`.
The same thing can be obtained using `reduction` OMP clause.

We used historically atomic because from what I read, reduction for arrays has been introduced in gcc6, previously only for single members.

This change increases performance, but also it makes code more readable, I think.

## Examples
image

---

### DIFF
--- a/Samples/SampleHandlerFD.cpp
+++ b/Samples/SampleHandlerFD.cpp
@@ -400,6 +400,8 @@ void SampleHandlerFD::FillArray() {
 }
 
 #ifdef MULTITHREAD
+#pragma GCC diagnostic push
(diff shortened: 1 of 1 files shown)


henry-wallace-phys (approved):
LGTM, are you able to get a step time comparison plot as well?

KSkwarczynski:
I was ultra-lazy...
Numbers in example are from running DragRace(1000)

Each number in the table is average over 1000 reweights.

This is also using tutorial

I think more interesting will be to see proper benchmark of T2K/DUNE one tag is done.

lazy-man excuse

PR #797, opened 2026-02-17
feat: Allow different Spline Class for FD

pgranger23:
# Pull request description
Tries to allow the use of different Spline classes for FD by changing the handler object type to `SplineBase`

## Changes or fixes
Changing SplineHandler to `SplineBase` type and use dynamic casting to keep the exact same behaviour as before when using `BinnedSplineHandler`

---

- [ x ] I have read and followed the [contributing guidelines](https://github.com/mach3-software/MaCh3/blob/develop/.github/CONTRIBUTING.md).


pgranger23:
Ok, clearly need to rework a bit the code as a few things have changed in develop since I wrote this code.

PR #815, opened 2026-03-04
breaking: Reduce number of pure virtual ReturnKinematicXXX functions

luketpickering:
make string versions of ReturnKinematicParameter final use the string->enum map to resolve themselves without subclass implementations.

Marked as final to stop overriding.

This will cause compilation failures for dependencies that override them.

# Pull request description

## Examples

```c++
  virtual double ReturnKinematicParameter(std::string KinematicParamter, int iEvent) final {
    return ReturnKinematicParameter(ReturnKinematicParameterFromString(KinematicParameter),iEvent);
  }
```

---


KSkwarczynski:
Looks good but will be merged once we start developing next relase with breaking changes

PR #823, opened 2026-03-07
tidy: SampleHandlerFD_data become vectors

luketpickering:
# Pull request description

```c++
SampleHandlerFD_array
SampleHandlerFD_array_w2
SampleHandlerFD_data
```

become `std::vector<double>`s, less manual memory management.

## Examples

---

### DIFF
--- a/Samples/SampleHandlerFD.cpp
+++ b/Samples/SampleHandlerFD.cpp
@@ -409,10 +409,10 @@ void SampleHandlerFD::FillArray_MP() {
   const auto TotalBins = Binning->GetNBins();
   const unsigned int NumberOfEvents = GetNEvents();
 
-  double *SampleHandlerFD_array_data = SampleHandlerFD_array.data();
-  double *SampleHandlerFD_array_w2_data = SampleHandlerFD_array_w2.data();
+  double *MC_Array_for_reduction = SampleHandlerFD_array.data();
+  double *W2_array_for_reduction = SampleHandlerFD_array_w2.data();
 
-  #pragma omp parallel for reduction(+:SampleHandlerFD_array_data[:TotalBins], SampleHandlerFD_array_w2_data[:TotalBins])
+  #pragma omp parallel for reduction(+:MC_Array_for_reduction[:TotalBins], W2_array_for_reduction[:TotalBins])
   for (unsigned int iEvent = 0; iEvent < NumberOfEvents; ++iEvent) {
     //ETA - generic functions to apply shifts to kinematic variables
     // Apply this before IsEventSelected is called.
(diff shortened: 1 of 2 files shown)


Samples/SampleHandlerFD.cpp
```diff
@@ -416,7 +408,11 @@ void SampleHandlerFD::FillArray_MP() {
   // We will hit <0.1 s/step eventually! :D
   const auto TotalBins = Binning->GetNBins();
   const unsigned int NumberOfEvents = GetNEvents();
-  #pragma omp parallel for reduction(+:SampleHandlerFD_array[:TotalBins], SampleHandlerFD_array_w2[:TotalBins])
+
+  double *SampleHandlerFD_array_data = SampleHandlerFD_array.data();
+  double *SampleHandlerFD_array_w2_data = SampleHandlerFD_array_w2.data();
+
```
KSkwarczynski: I know why you called it _data but it is bit overloaded term in this context

Samples/SampleHandlerFD.cpp
```diff
@@ -1282,9 +1270,9 @@ double SampleHandlerFD::GetSampleLikelihood(const int isample) const {
   #endif
   for (int idx = Start; idx < End; ++idx)
   {
-    const double DataVal = SampleHandlerFD_data[idx];
-    const double MCPred = SampleHandlerFD_array[idx];
-    const double w2 = SampleHandlerFD_array_w2[idx];
+    double const &DataVal = SampleHandlerFD_data[idx];
+    double const &MCPred = SampleHandlerFD_array[idx];
+    double const &w2 = SampleHandlerFD_array_w2[idx];
```
KSkwarczynski, line 1275: Why reference? Is it faster?
We do not want to modify them?

KSkwarczynski (approved):
LGTM other than minor comments

Samples/SampleHandlerFD.cpp
luketpickering: just because i need a pointer/carray rather than a vector for the omp and the way i get that is calling .data on the vector... i agree data is overloaded. maybe even 'for_reduction' or something longer but explicit?

Samples/SampleHandlerFD.cpp
luketpickering: a reference is a pointer under the hood, it is smaller than a double. its possible these doubles are kept in a register somewhere but probable that if its not a reference they are copied. its definitely premature optimization to make these a reference, but it also doesnt hurt to basically always use const references when you can.

KSkwarczynski: I am all for premature optimization ;) 
Just wanted to understand thought process

Samples/SampleHandlerFD.cpp
KSkwarczynski: You can also do `MC_Array_for_reduction` and `W2_array_for_reduction` as well, I think.
Don't have strong preference other than not use `_data`

PR #824, opened 2026-03-07
breaking: Multi-parameter functional shifts

luketpickering:
# Pull request description
Attempts to design new Functional Shift interface. Currently, but could probably be made not breaking.

For more see: https://github.com/mach3-software/MaCh3Tutorial/pull/254
## Changes or fixes
Key design change is that functional shifts can now be functions of multiple parameters and operate on experiment class event types, rather than taking event indexes. The experiment class event vector is now passed as the first parameter.

The hope was to remove any capturing lambdas from the tight event loop, this implementation doesn't achieve that yet.

## Examples

```c++
  RegisterIndividualFunctionalParameter(
      DUNEMCEvents, "MyParameter", [](double const *par, EventInfo &ev) {
        std::cout << "I'm here with " << (*par) << std::endl;
        ev.varied_reco.e_lep *= (1.0 + *par);
      });

  RegisterIndividualFunctionalParameter(
      DUNEMCEvents, std::vector<std::string>{"MyParameter", "MyOtherParameter"},
      [](std::vector<double const *> const &pars, EventInfo &ev) {
        std::cout << "I'm here with " << (*pars[0]) << " and " << (*pars[1])
                  << std::endl;
        ev.varied_reco.e_lep *= (1.0 + *pars[0] + *pars[1]);
      });
```

---


luketpickering:
Slight improvement to the proposed sampleHandler interface shown below:

```c++
  RegisterIndividualFunctionalParameter(
      DUNEMCEvents, "MyParameter", [](double const &par, EventInfo &ev) {
        std::cout << "I'm here with " << par << std::endl;
        ev.varied_reco.e_lep *= (1.0 + par);
      });

  RegisterIndividualFunctionalParameter(
      DUNEMCEvents, std::vector<std::string>{"MyParameter", "MyOtherParameter"},
      [](std::vector<double> const &pars, EventInfo &ev) {
        std::cout << "I'm here with " << pars[0] << " and " << pars[1]
                  << std::endl;
        ev.varied_reco.e_lep *= (1.0 + pars[0] + pars[1]);
      });
```

Now the relevant parameter values are fetched and cached before the event loop, this allows us to pass a vector of parameter values instead of a vector of pointers to parameter values which makes usage cleaner and some patterns more flexible.

KSkwarczynski (approved):
LGTM

PR #829, opened 2026-03-12
feat: Input file use in GetStepScaleBasedOnLLHScan 

menailj:
# Pull request description
Attempting to use GetStepScaleBasedOnLLHScan and run both with and without an input file with an LLH scan specified. 

## Changes or fixes
Ensure file with llh scans exists to stop code breaking. Then allow user to specify their own input file rather than having to run llh scan each time the code is called. 

## Examples

---


Fitters/FitterBase.cpp
```diff
@@ -929,6 +935,7 @@ void FitterBase::GetStepScaleBasedOnLLHScan() {
     cov->SetIndivStepScale(StepScale);
     cov->SaveUpdatedMatrixConfig();
   }
+  delete outputFileLLH;
 }
```
KSkwarczynski: outputFileLLH is by default nullptr.

it will be much safer to do
`if(outputFileLLH != nullptr)  delete outputFileLLH;`

Fitters/FitterBase.cpp
```diff
@@ -884,25 +883,32 @@ void FitterBase::RunLLHScan() {
...
 // *************************
 //LLH scan is good first estimate of step scale
-void FitterBase::GetStepScaleBasedOnLLHScan() {
+void FitterBase::GetStepScaleBasedOnLLHScan(const std::string& outputFileName) {
 // *************************
-  TDirectory *Sample_LLH = outputFile->Get<TDirectory>("Sample_LLH");
+  TFile* outputFileLLH = NULL;
+  if(outputFileName != NULL){
+    outputFileLLH = M3::Open(outputFileName, "READ", __FILE__, __LINE__);
+  }
```
KSkwarczynski: std::string can't be `nullptr`.

It would be safer to do
`if(outputFileName != ""){`
and in header of `FitterBase `add default argument

KSkwarczynski (changes requested):

KSkwarczynski (approved):
LGTM

PR #841, opened 2026-03-26
breaking: NuDock to work in debug mode

KSkwarczynski:
# Pull request description
NuDock is using variable DEBUG. MaCh3 in debug mode was exporting DEBUG as definition...
Which is terrible design.
Now `ifdef DEBUG` was renamed to `ifdef MACH3_DEBUG`. This allows for NuDock to compile in debug mode.

---


Parameters/PCAHandler.h
```diff
@@ -4,11 +4,11 @@
 #include "Manager/Manager.h"
 #include "Parameters/ParameterHandlerUtils.h"
 
-#ifdef DEBUG
+#ifdef MACH3_DEBUG
```
ewanwm: since MACH3_DEBUG_PCA only gets set if MACH3_DEBUG is set can we just get rid of MACH3_DEBUG_PCA and replace it with MACH3_DEBUG?

ewanwm (approved):
minor but looks good!

PR #851, opened 2026-03-31
bugfix: SampleHandlerBase::Selection is left empty by initialize

luketpickering:
- Currently the 'main' Selection vector is only set the first time FillArray is called, which means that IsEventSelected cannot be used before FillArray is used (e.g. to interrogate spline applicability.)

# Pull request description

## Examples

---


KSkwarczynski (approved):
LGTM

PR #855, opened 2026-04-09
breaking: Improve cuts based on SampleName

KSkwarczynski:
# Pull request description
MaCh3 had the ability to apply cuts based on sample name. However, in past, sample name was same as SampleHandler Name. Once we allowed to have multiple sample in one object, cut was being done based on `SampleHandlerName`.
This isn't sufficient for analysis like T2K+SK where they need to apply cuts within the same object.

Now cuts apply based on `SamleName `(individual). Thus, a new field is required
`SampleHandlerName `-> Name of object
`SampleTitle `-> Simply fancy name
`SampleName `-> Tag used to define by which syst to be affected

By default, if `SampleName `is not defined it will use `SampleHandlerName `which should reduce number of changes for people not needing this feature.

---

### DIFF
--- a/Samples/SampleHandlerBase.cpp
+++ b/Samples/SampleHandlerBase.cpp
@@ -1046,7 +1046,7 @@ void SampleHandlerBase::InitialiseNuOscillatorObjects() {
     }
   }
   // get osc params for sample 0, later we check all have same number
-  std::vector<const M3::float_t*> OscParams = ParHandler->GetOscParsFromSampleName(0);
+  std::vector<const M3::float_t*> OscParams = ParHandler->GetOscParsFromSampleName(GetSampleName(0));
   if (OscParams.empty()) {
     MACH3LOG_ERROR("OscParams is empty for sample '{}'.", GetName());
     MACH3LOG_ERROR("This likely indicates an error in your oscillation YAML configuration.");
--- a/Samples/SampleHandlerBase.cpp
+++ b/Samples/SampleHandlerBase.cpp
@@ -110,6 +110,8 @@ void SampleHandlerBase::LoadSingleSample(const int iSample, const YAML::Node& Sa
   SampleInfo SingleSample;
   //SampleTitle has to be provided in the sample yaml otherwise this will throw an exception
   SingleSample.SampleTitle = Get<std::string>(SampleSettings["SampleTitle"], __FILE__ , __LINE__);
+  // Sample name used for defying syst uncertainties, if not specified use SampleHandler name
+  SingleSample.SampleName = GetFromManager<std::string>(SampleSettings["SampleName"], GetName(), __FILE__ , __LINE__);
 
   Binning->SetupSampleBinning(SampleSettings["Binning"], SingleSample);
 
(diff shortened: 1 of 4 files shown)


Samples/SampleHandlerBase.cpp
```diff
@@ -1049,12 +1045,22 @@ void SampleHandlerBase::InitialiseNuOscillatorObjects() {
...
   if (OscParams.empty()) {
     MACH3LOG_ERROR("OscParams is empty for sample '{}'.", GetName());
     MACH3LOG_ERROR("This likely indicates an error in your oscillation YAML configuration.");
     throw MaCh3Exception(__FILE__, __LINE__);
   }
+
+  for(int iSample = 1; iSample < GetNSamples(); iSample++) {
+    auto OscParamsTest = ParHandler->GetOscParsFromSampleName(GetSampleName(iSample));
+    if (OscParamsTest.size() != OscParams.size()) {
+      MACH3LOG_ERROR("Sammple {} has {} osc params while sample {} has {}",
```
dbarrow257: Typo

Samples/SampleHandlerBase.cpp
```diff
@@ -1049,12 +1045,22 @@ void SampleHandlerBase::InitialiseNuOscillatorObjects() {
...
+  // get osc params for sample 0, later we check all have same number
+  std::vector<const M3::float_t*> OscParams = ParHandler->GetOscParsFromSampleName(GetSampleName(0));
   if (OscParams.empty()) {
     MACH3LOG_ERROR("OscParams is empty for sample '{}'.", GetName());
     MACH3LOG_ERROR("This likely indicates an error in your oscillation YAML configuration.");
     throw MaCh3Exception(__FILE__, __LINE__);
   }
+
+  for(int iSample = 1; iSample < GetNSamples(); iSample++) {
+    auto OscParamsTest = ParHandler->GetOscParsFromSampleName(GetSampleName(iSample));
```
dbarrow257: Test isn't a super descriptive variable name suffix

Samples/SampleHandlerBase.h
```diff
@@ -281,7 +282,7 @@ class SampleHandlerBase :  public SampleHandlerInterface
   std::vector<std::string> funcParsNamesVec = {};
 
   /// @brief Check whether a normalisation systematic affects an event or not
-  void CalcNormsBins(std::vector<NormParameter>& norm_parameters, std::vector< std::vector< int > >& norms_bins);
+  void CalcNormsBins(std::vector<std::vector<NormParameter>>& norm_parameters, std::vector< std::vector< int > >& norms_bins);
```
dbarrow257, line 286: Might be worth adding what the vector indexing is in the comments

Samples/SampleHandlerBase.cpp
```diff
@@ -591,7 +600,7 @@ void SampleHandlerBase::SetupOscParameters() {
   // KS: Only make sense to setup osc if you have ParHandler
   if(ParHandler == nullptr ) return;
 
-  auto OscParams = ParHandler->GetOscParsFromSampleName(SampleHandlerName);
+  auto OscParams = ParHandler->GetOscParsFromSampleName(GetSampleName(0));
```
dbarrow257, line 608: Why is this (0) - Maybe I don't quite follow the logic

dbarrow257: It looks like you added a test to ensure all samples have the same osc pars - so fine. But what happens if there are no samples? I assume that would be caught somewhere else?

dbarrow257 (approved):
LGTM - How does this change how the cov mat definition?

Samples/SampleHandlerBase.cpp
KSkwarczynski: Thanks

Samples/SampleHandlerBase.cpp
KSkwarczynski: Changed to OscParamsCrossCheck

Samples/SampleHandlerBase.h
KSkwarczynski: Added doxygen comment

Samples/SampleHandlerBase.cpp
KSkwarczynski: All samples within one sample handler should have same oscillation physics. No mixing ATM/Beam samples.

And now I added check that nSamples can't be 0

PR #877, opened 2026-04-24
tidy: Remove memory leaks (minor) in PSO code

KSkwarczynski:
# Pull request description
As title says, simply use modern C++ instead of raw pointers

## Examples

---

### DIFF
--- a/Fitters/PSO.cpp
+++ b/Fitters/PSO.cpp
@@ -33,9 +33,9 @@ void PSO::RunMCMC(){
   if(fTestLikelihood){
     outTree->Branch("nParts", &fParticles, "nParts/I");
     for(int i = 0; i < fDim; ++i){
-      par = new double[fParticles];
-      paramlist.push_back(par);
-      outTree->Branch(Form("Parameter_%d", i), paramlist[i], Form("Parameter_%d[nParts]/D",i));
+      std::vector<double> par(fParticles);
+      paramlist.push_back(std::move(par));
+      outTree->Branch(Form("Parameter_%d", i), paramlist[i].data(), Form("Parameter_%d[nParts]/D",i));
     }
 //  vel = new double[fParticles];
     outTree->Branch("vel", vel, "vel[nParts]/D");
@@ -135,15 +135,15 @@ void PSO::init(){
       }
     }
 
-    particle* new_particle = new particle(init_position,init_velocity);
+    auto new_particle = std::make_unique<particle>(init_position,init_velocity);
     new_particle->set_personal_best_position(init_position);
     double new_value = CalcChi(init_position);
     new_particle->set_personal_best_value(new_value);
     new_particle->set_value(new_value);
-    system.push_back(new_particle);
+    system.push_back(std::move(new_particle));
     if(new_value < fBestValue){
       fBestValue = new_value;
-      set_best_particle(new_particle);
+      set_best_particle(system.back().get());
     }
   }
 }
@@ -322,7 +322,7 @@ void PSO::uncertainty_check(const std::vector<double>& previous_pos){
     position[i] = curr_ival;
     MACH3LOG_INFO("");
     MACH3LOG_INFO("For fDim{} x list is", i+1);
-    for (unsigned int k = 0;k<x.size(); ++k){
+    for (unsigned int k = 0; k < x.size(); ++k){
       MACH3LOG_INFO("  {}", x[k]);
     }
     MACH3LOG_INFO("");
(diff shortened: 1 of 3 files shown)


Fitters/PSO.cpp
```diff
@@ -376,7 +376,7 @@ double PSO::swarmIterate(){
       system[i]->set_personal_best_position(new_pos);
       if(new_value < fBestValue){
         fBestValue = new_value;
-        set_best_particle(system[i]);
+        set_best_particle(system[i].get());
```
henry-wallace-phys, line 377: If we're having to do gets, does it not make sense to have std::shared_ptr<...> instead?

Fitters/PSO.cpp
```diff
@@ -464,54 +464,54 @@ void PSO::WriteOutput(){
     }
   }
   else{
-    for (std::vector<ParameterHandlerBase*>::iterator it = systematics.begin(); it != systematics.end(); ++it)
+    for (std::vector<ParameterHandlerBase*>::iterator ParHandler = systematics.begin(); ParHandler != systematics.end(); ++ParHandler)
```
henry-wallace-phys: Wait we can just do an (auto& par: systematics){...} here can't we rather than using std::iterators?

henry-wallace-phys (approved):
Lots of nitpicks, approving since it "works" but would be good to fix

Fitters/PSO.cpp
luketpickering: No, pls don't use shared pointer, every deref requires a mutex for the ref counter.

You can pass references of unique pointers, you just can't copy them.

Fitters/PSO.cpp
```diff
@@ -33,9 +33,9 @@ void PSO::RunMCMC(){
   if(fTestLikelihood){
     outTree->Branch("nParts", &fParticles, "nParts/I");
     for(int i = 0; i < fDim; ++i){
-      par = new double[fParticles];
-      paramlist.push_back(par);
-      outTree->Branch(Form("Parameter_%d", i), paramlist[i], Form("Parameter_%d[nParts]/D",i));
+      std::vector<double> par(fParticles);
+      paramlist.push_back(std::move(par));
```
luketpickering: you don't need the temporary if you immediately move it, replace with:

```c++
paramlist.emplace_back(fParticles);
```

Fitters/PSO.cpp
luketpickering: or at least deduce the horrible iterator type with `auto`.

Fitters/PSO.cpp
```diff
@@ -464,54 +464,54 @@ void PSO::WriteOutput(){
...
   }
   else{
-    for (std::vector<ParameterHandlerBase*>::iterator it = systematics.begin(); it != systematics.end(); ++it)
+    for (std::vector<ParameterHandlerBase*>::iterator ParHandler = systematics.begin(); ParHandler != systematics.end(); ++ParHandler)
     {
-      if(!(*it)->IsPCA())
+      if(!(*ParHandler)->IsPCA())
       {
-        for(int i = 0; i < (*it)->GetNumParams(); ++i, ++ParCounter)
+        for(int i = 0; i < (*ParHandler)->GetNumParams(); ++i, ++ParCounter)
```
luketpickering: Henry's comment about use range-based for would make this neater too.. as then it becomes `ParHandler->GetNumParams()`

Fitters/PSO.cpp
```diff
@@ -135,15 +135,15 @@ void PSO::init(){
...
     }
 
-    particle* new_particle = new particle(init_position,init_velocity);
+    auto new_particle = std::make_unique<particle>(init_position,init_velocity);
     new_particle->set_personal_best_position(init_position);
     double new_value = CalcChi(init_position);
     new_particle->set_personal_best_value(new_value);
     new_particle->set_value(new_value);
-    system.push_back(new_particle);
+    system.push_back(std::move(new_particle));
```
luketpickering: I think its worth being smidge more careful with `std::move` and not inviting use-after-move:

```
system.emplace_back(new particle(init_position,init_velocity));
auto &new_particle = system.back(); new_particle->set_personal_best_position(init_position);
double new_value = CalcChi(init_position);
new_particle->set_personal_best_value(new_value);
```

Fitters/PSO.cpp
luketpickering: Maybe shared_ptr is fine here, it depends how tight/multithreaded this code is

luketpickering: Looking at the code the vector of particles once initialized is not resized, so this could use the particle index as the unique identifier rather than a pointer at all.

```set_best_particle(i);```

and

```system[get_best_particle()]->bla()```

luketpickering (changes requested):

PR #903, opened 2026-05-13
tidy: Remove Duplicates in ReturnHistsBySelection

KSkwarczynski:
# Pull request description
Simply, we no longer define twice getter for iMax.
It is simple code but since we declare it twice better to have helper function

## Changes or fixes
- Check bounds now uses branchless comparison instead of if, and without multithreading. Did benchmark and new is much faster. It doesn't matter at all as this operation is sub percent of total step time...
- Renamed FDPlotType into SamplePlotType

## Examples

Benchmark results for check bounds
```
===== N = 100 =====
Branchless: 0.021882 ms (result=48000)
Scalar: 0.053182 ms (result=48000)
OMP Branchless: 1.76174 ms (result=48000)
OMP: 1.06407 ms (result=48000)

===== N = 1000 =====
Branchless: 0.347344 ms (result=466000)
Scalar: 1.26515 ms (result=466000)
OMP Branchless: 1.08954 ms (result=466000)
OMP: 1.03105 ms (result=466000)

===== N = 5000 =====
Branchless: 1.79323 ms (result=2461000)
Scalar: 4.45375 ms (result=2461000)
OMP Branchless: 1.3847 ms (result=2461000)
OMP: 1.8341 ms (result=2461000)
[root@be08be133572 hosthome]#
```

---

### DIFF
--- a/Samples/SampleHandlerBase.cpp
+++ b/Samples/SampleHandlerBase.cpp
@@ -1991,7 +1991,7 @@ int SampleHandlerBase::GetRangeForPlotType(const SamplePlotType TypeEnum, const 
       return GetNOscChannels(iSample);
       break;
     default:
-      MACH3LOG_ERROR("You've passed me a Selection1 which was not implemented in ReturnHistsBySelection1D. Selection1 and Selection2 are counters for different indexable quantities");
+      MACH3LOG_ERROR("You've passed me a SamplePlotType with value {} which was not implemented.", TypeEnum);
       throw MaCh3Exception(__FILE__, __LINE__);
   }
 }
--- a/Samples/SampleHandlerBase.cpp
+++ b/Samples/SampleHandlerBase.cpp
@@ -1870,9 +1870,10 @@ void SampleHandlerBase::PrintIntegral(const int iSample, const TString& OutputFi
 
   for (int i=0;i<Modes->GetNModes();i++) {
     if (GetNDim(iSample) == 1) {
-      IntegralList[i] = ReturnHistsBySelection1D(iSample, GetKinVarName(iSample, 0),1,i,WeightStyle);
+      IntegralList[i] = ReturnHistsBySelection1D(iSample, GetKinVarName(iSample, 0), SamplePlotType::kOscChannelPlot, i, WeightStyle);
     } else {
-      IntegralList[i] = CastVector<TH2, TH1>(ReturnHistsBySelection2D(iSample, GetKinVarName(iSample, 0), GetKinVarName(iSample, 1),1,i,WeightStyle));
+      IntegralList[i] = CastVector<TH2, TH1>(ReturnHistsBySelection2D(iSample, GetKinVarName(iSample, 0), GetKinVarName(iSample, 1),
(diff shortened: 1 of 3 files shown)


Samples/SampleHandlerBase.cpp
```diff
@@ -1979,6 +1978,24 @@ void SampleHandlerBase::PrintIntegral(const int iSample, const TString& OutputFi
   }
 }
 
+// ************************************************
+int SampleHandlerBase::GetRangeForPlotType(const int TypeEnum, const int iSample) const {
```
henry-wallace-phys: nitpick/style choice, is there a reason for not just doing `return` in each case rather than assigning iMax at the start?

Samples/SampleHandlerBase.h
```diff
@@ -108,11 +109,14 @@ class SampleHandlerBase :  public SampleHandlerInterface
   std::unique_ptr<TH2> Get2DVarHist(const int iSample, const std::string& ProjectionVarX, const std::string& ProjectionVarY,
                                     const std::vector< KinematicCut >& EventSelectionVec = {},
                                     int WeightStyle = 0, const std::vector< KinematicCut >& SubEventSelectionVec = {}) final;
+  /// @brief Construct vector of kinematic cuts that will be applied, on top of default cuts include stuff like cut on mode etc.
+  /// @param Sample Index of the sample.
   std::vector<KinematicCut> BuildModeChannelSelection(const int iSample, const int kModeToFill, const int kChannelToFill) const;
-
+  /// @brief Fill projection histogram by looping over all events, and skipping one which doens't pass specified condition
```
henry-wallace-phys: typo (doens't -> doesn't)

henry-wallace-phys (changes requested):
Small nitpicks. One major comment, rather than raw `const int` calls, can we make these selections (`Selection1`, `Selection2`) use `Enums` instead to avoid magic numbers?

```
  std::vector<std::unique_ptr<TH1>> ReturnHistsBySelection1D(const int iSample, const std::string& KinematicProjection, const int Selection1, const int Selection2 = -1,...
```

Samples/SampleHandlerBase.cpp
```diff
@@ -1979,6 +1978,24 @@ void SampleHandlerBase::PrintIntegral(const int iSample, const TString& OutputFi
...
+  int iMax = -1;
+  switch (TypeEnum) {
+    case SamplePlotType::kModePlot:
+      iMax = Modes->GetNModes();
+      break;
+    case SamplePlotType::kOscChannelPlot:
+      iMax = GetNOscChannels(iSample);
+      break;
+    default:
+      MACH3LOG_ERROR("You've passed me a Selection1 which was not implemented in ReturnHistsBySelection1D. Selection1 and Selection2 are counters for different indexable quantities");
```
henry-wallace-phys: Can we make this more helpful? this error doesn't really tell me what I'm allow to do

henry-wallace-phys (approved):
LGTM, other request (const int->direct enum call) results in breaking changes and will be addressed in a future patch

PR #904, opened 2026-05-13
tidy: add deprecation warning for GetModeHist1D

KSkwarczynski:
# Pull request description
For more see this: 
https://t2k-experiment.slack.com/archives/C08MMSCFWDV/p1778667053954489

but tl:dr these functions are not used and and hardcoded to assume X-Var/Y-Var which can be wrong for some analysis in future.

## Examples
image

---


BenJarg (approved):
LGTM - I don't see it in use in T2K or DUNE MaCh3's currently, and this won't actually break any code using it somewhere else anyway

PR #907, opened 2026-05-15
tidy: cmake improvments

KSkwarczynski:
# Pull request description
Several of cmake improvements

## Changes or fixes
- NuOsc version is not added to version.h
- Libraries are added to MaCh3 all in individual cmake. If we reach a point where MaCh3 has more directories it will make easier to sustain it.
- In find_package ROOT requires same version as in compilation, not any version larger than 6.20. This would not work most likely and cause difficult to debug problems.
- Multiple directories had CUDA architecture even if they didn't compile any GPU code...

## Examples
image

---


CMakeLists.txt
```diff
@@ -3,7 +3,7 @@ set(CMAKE_VERBOSE_MAKEFILE ON)
 # CMake version check
 cmake_minimum_required(VERSION 3.14 FATAL_ERROR)
 
-project(MaCh3 VERSION 2.5.1 LANGUAGES CXX)
+project(MaCh3 VERSION 2.6.0 LANGUAGES CXX)
```
henry-wallace-phys, line 6: is this PR targeting 2.6.0?

KSkwarczynski: We just tagged 2.5.1 so might as well bump version to next one.
Dan, Luke and You hinted interest into features with breaking changes

PR #909, opened 2026-05-18
feat: Sanitizer for adaption and flipping

KSkwarczynski:
# Pull request description
We do know adaption doesn't work very well with flipping.
So let's add sanitizer.

If this is too harsh, can change error to warning.
## Changes or fixes
- Change xsec_cov into ParamHandler in Sample Handler

## Examples
image

---


Parameters/ParameterHandlerBase.cpp
```diff
@@ -961,6 +961,39 @@ void ParameterHandlerBase::UpdateThrowMatrix(TMatrixDSym *cov) {
...
+// ********************************************
+void ParameterHandlerBase::SanitizeAdaption() const {
+// ********************************************
+  for (size_t i = 0; i < FlipParameterIndex.size(); ++i) {
+    const int index = FlipParameterIndex[i];
+    if(!param_skip_adapt_flags[index]) {
+      MACH3LOG_ERROR("You enabled adaption for parameter which has enabled flipping ({})", _fFancyNames[index]);
+      MACH3LOG_ERROR("Right now flipping and adapting doesn't work very well");
+      MACH3LOG_ERROR("Please skip adaption for param {}, using ParametersToSkip option in config", _fFancyNames[index]);
+      throw MaCh3Exception(__FILE__, __LINE__);
```
henry-wallace-phys, line 974: So I think throwing an exception here is maybe a bit harsh, is it possible to pass it an override flag at setup? 

Something like this in the YAML:
```
SanitizeAdaption: <True/False>
```
that has default behaviour of `True`. The default (True) will run `ParamaterHandlerBase::SanitizeAdaption` as is whilst `False` does something like

 `MACH3LOG_CRITICAL("You've turned of sanity checking adaption! Be incredibly careful, this will not check for correlated parameter adaptions and will allow flipped parameters to adapt without issue")` ?

henry-wallace-phys (changes requested):
I think the current sanitiser needs an override for "wacky" behaviours but otherwise looks great

henry-wallace-phys (approved):
After discussion with Kamil decided override is overkill + can be added in future

PR #917, opened 2026-05-28
feat: Unified binary (CLI)

Charlotte-Knight:
## User Interface

This PR introduces a single binary through which to run MaCh3 executables. In other words, it introduces a unified CLI. 

For example, one can run `mach3 --help` and get:
```
Usage:  [--help] [--version] {diag,penterm,process,tutorial}

Optional arguments:
  -h, --help     shows help message and exits
  -v, --version  prints version information and exits

Subcommands:
  diag
  penterm
  process       Main exectable responsible for different types of MCMC processing like drawing posteriors, triangle plots etc.
```
and then run `ProcessMCMC` with:
```
mach3 process config.yaml chain.root
```

So far, the following executables have been implemented:
- DiagMCMC
- ProcessMCMC
- GetPenaltyTerm

And the functionality exists for experiments to add their own executables via plugins. An example of this is included in MaCh3Tutorial via this PR: https://github.com/mach3-software/MaCh3Tutorial/pull/252. There, you can run the `MCMCTutorial` executable like:
```
mach3 tutorial TutorialConfigs/FitterConfig.yaml --MCMCSteps 100000
```

The previous name for the commands can still be used like:
```
ProcessMCMC config.yaml chain.root
```
and the user will be presented with a deprecation warning like:
```
[2026-06-01 14:04:37.661] [warning] [ProcessMCMC.cpp:5] Deprecation Warning: Use of the standalone executable will be deprecated in future releases.
[2026-06-01 14:04:37.661] [warning] [ProcessMCMC.cpp:6]                    : you can use 'mach3 process' as a direct replacement instead.
```

### Feedback
We're open to suggestions about the naming of the commands, e.g. whether `mach3 process` makes sense or maybe `mach3 processmcmc` etc. And anything other comments are welcome of course.

## Future Plans

Assuming that this PR is merged, we imagine that this executable will start to gain traction with some users, and hopefully, we can gather feedback. In time, we can implement that feedback and eventually implement the rest of the executables and remove the old interface.

## Implementation 

TODO: Comments about implementation for the other devs. For example, all of the executables going under the CLI/API directory.

## Subcommand map

| Old Executable Name  | CLI Subcommand |
| ------------- | ------------- |
| ProcessMCMC  | process  |
| DiagMCMC | diag  |
| GetPenaltyTerm | penterm |

---


CLI/API/CMakeLists.txt
```diff
@@ -0,0 +1,35 @@
+include(FetchContent)
+FetchContent_Declare(
+    argparse
+    GIT_REPOSITORY https://github.com/p-ranav/argparse.git
+)
+FetchContent_MakeAvailable(argparse)
```
KSkwarczynski, line 7: We should use tag or commit rather than master?

For fetching, we use CPM. I don't have strong opinion for CPM vs FetchContent.
Part of me thinks nice to use same for all as it is less confusing, but doesn't matter a lot.

Might be worth considering moving to: https://github.com/mach3-software/MaCh3/blob/develop/cmake/Modules/MaCh3Dependencies.cmake
but again don't have strong opinion on this one.

cli/modules/DiagMCMCModule.cpp
```diff
@@ -0,0 +1,57 @@
+#include "Fitters/MCMCProcessor.h"
+#include "Manager/Manager.h"
+#include "cli/modules/DiagMCMCModule.hpp"
```
KSkwarczynski: We use capital letters for most folders. Could we use capital for CLI related?

cli/api/argparse.hpp
```diff
@@ -0,0 +1,29 @@
...
+
+            const std::string name() const{
+                return this->m_program_name;
+            }
+            const std::list<std::reference_wrapper<ArgumentParser>>& subparsers() const{
+                return this->m_subparsers;
+            }
+
+            const MaCh3ArgumentParser& get_subcommand_used() const{
+                for (const std::reference_wrapper<ArgumentParser>& subparser : this->m_subparsers) {
```
KSkwarczynski, line 41: I am making comment here but it relates to other parts of this PR, namely no Doxygen comments.

Don't need super detailed description but simple this class does XYZ is needed I think.

You also add /// @author if you like

cli/api/plugin.hpp
```diff
@@ -0,0 +1,30 @@
...
+}
+
+#define MACH3_REGISTER_PLUGIN(PluginClass)                     \
+static_assert(std::is_base_of<M3::IPlugin, PluginClass>::value, "PluginClass must derive from M3::IPlugin");  \
+extern "C" M3::IPlugin* create_plugin() {                   \
+    return new PluginClass();                                  \
+}                                                              \
+extern "C" void destroy_plugin(M3::IPlugin* p) {            \
+    delete p;                                                  \
+}
```
KSkwarczynski: I am almost certainly missing something, but is there strong reason to use raw pointers instead of smart?

Probably Doxygen comments would help in future confusion

CLI/Modules/DiagMCMCModule.cpp
```diff
@@ -0,0 +1,57 @@
...
+namespace M3{
+  
+  DiagMCMCModule::~DiagMCMCModule(){
+    if (this->m_parser) { delete this->m_parser; } 
+  }
+
+  MaCh3ArgumentParser* DiagMCMCModule::get_parser(){
+    m_parser = new MaCh3ArgumentParser("diag", "1.0", argparse::default_arguments::help);
+    m_parser->add_argument("mcmc-output")
+      .help("MCMC chain root file.")
```
KSkwarczynski, line 19: I understand 1.0 is a version of parser, but is it because `ArgumentParser `require it, or this is standard?
Should we bump version everytime we modify the parser?

Can we not include version?

I think some Doxygen comments here and there would help in future maintenance etc.

cli/api/plugin.hpp
```diff
@@ -0,0 +1,30 @@
...
+#include "cli/api/argparse.hpp"
+
+
+namespace M3 {
+
+    class IPlugin {
+        public:
+            virtual ~IPlugin() = default;
+            virtual int run() = 0;
+            virtual MaCh3ArgumentParser* get_parser() = 0;
```
KSkwarczynski: Purely preference but can we use Run instead of run?

cli/modules/DiagMCMCModule.hpp
```diff
@@ -0,0 +1,17 @@
...
+    public:
+      virtual ~DiagMCMCModule();
+      MaCh3ArgumentParser* get_parser() override;
+      int run() override;
+
+
+    private:
+      MaCh3ArgumentParser* m_parser;
+  };
+};
```
KSkwarczynski: This PR stripped parts of Doxygen we had, like this
https://github.com/mach3-software/MaCh3/blob/d491feb7a2f59861baf81683292c958ebcacf873/Diagnostics/DiagMCMC.cpp#L5-L14

I have quite strong opinion about including @author for contributions that MaCh3 received over the years.

I am not saying documenataion need to be identical but we should not strip authors.

cli/modules/GetPenaltyTermModule.cpp
```diff
@@ -0,0 +1,410 @@
...
+/// @file GetPenaltyTerm.cpp
+/// @brief KS: This file contains the implementation of the function to extract specific penalty terms from systematic chains.
+///
+/// This script is designed to retrieve penalty terms from various sources, such as flux and cross-section systematic chains.
+/// Since flux and cross-section uncertainties are handled systematically, the penalty term cannot be taken directly from the chain.
+/// @todo KS: This should really be moved to MCMC Processor
+///
+/// @ingroup MaCh3DiagnosticProcessing
+///
+/// @author Kamil Skwarczynski
```
KSkwarczynski: This probably can be moved now to header file as it is not class not single file

CLI/Modules/ProcessMCMCModule.cpp
```diff
@@ -0,0 +1,853 @@
...
+      .flag();
+    m_parser->add_argument("config")
+      .help("Config file.")
+      .metavar("CONFIG")
+      .required();
+    m_parser->add_argument("mcmc-chain")
+      .help("MCMC chain root files and titles.\n"
+            "single chain mode: MCMC_CHAIN1\n"
+            "two chain mode   : MCMC_CHAIN1 TITLE1 MCMC_CHAIN2 TITLE2\n"
+            "three chain mode : MCMC_CHAIN1 TITLE1 MCMC_CHAIN2 TITLE2 MCMC_CHAIN3 TITLE3")
```
KSkwarczynski, line 55: **Not for this PR**, but would be neat to pass a few chains, which are added as TChain

`For example mcmc process --mcm-chain chain1_setting1.root;chain2_setting1.root --mcm-chain chain1_setting2.root;chain2_setting2.root`

CLI/MaCh3Program.cpp
```diff
@@ -0,0 +1,232 @@
...
+        if (s.find("zsh")  != std::string::npos) return "zsh";
+
+        return "";
+    }
+
+    const void MaCh3Program::install_completions() const{
+        const char* home = std::getenv("HOME");
+        if (!home) {
+            std::cerr << "Cannot determine HOME directory\n";
+            return;
```
KSkwarczynski, line 44: Is std::cerr etc used here because it is part of CLI not actual executable?

cli/api/argparse.hpp
```diff
@@ -0,0 +1,29 @@
+#pragma once
+#include <argparse/argparse.hpp>
+
```
KSkwarczynski, line 6: Can we hide this under _MaCh3_Safe_Include_Start_   ?

https://github.com/mach3-software/MaCh3/blob/d491feb7a2f59861baf81683292c958ebcacf873/Manager/Core.h#L122-L125

CLI/mach3.cpp
```diff
@@ -0,0 +1,69 @@
+#include "cli/MaCh3Program.hpp"
+#include "cli/modules/ProcessMCMCModule.hpp"
+#include "cli/modules/DiagMCMCModule.hpp"
+#include "cli/modules/GetPenaltyTermModule.hpp"
+
+
+int main(int argc, char *argv[]) {
+    M3::MaCh3Program program("mach3");
+
+    // Hidden completion option
```
KSkwarczynski: What --complete do?

I know naive question but I think some comment would help

CMakeLists.txt
```diff
@@ -305,7 +307,12 @@ add_library(MaCh3 INTERFACE)
 if (MaCh3_NUDOCK_ENABLED)
   target_link_libraries(MaCh3 INTERFACE MaCh3NuDock)
 endif()
-target_link_libraries(MaCh3 INTERFACE Fitters Samples Parameters Splines Manager MaCh3CompilerOptions MaCh3GPUCompilerOptions Plotting)
+target_link_libraries(MaCh3 INTERFACE CLIApi Fitters Samples Parameters Splines Manager MaCh3CompilerOptions MaCh3GPUCompilerOptions Plotting)
+# Ensure API headers propagate to all consumers of MaCh3::All
+# target_include_directories(MaCh3 INTERFACE
+#     $<BUILD_INTERFACE:${CMAKE_CURRENT_SOURCE_DIR}/api>
+#     $<INSTALL_INTERFACE:include/api>
+# )
```
KSkwarczynski: This is example but there is plenty of commented code from testing.

Just marking as we should ideally clean PR before opening

KSkwarczynski (commented):
I have plenty of comments, although majority are nitpicky and style.

My main comment is lack of documentation or short explanation why X was done like this etc.

Doesn't have to be encyplopedia.

KSkwarczynski:
I also think this need to be updated
https://github.com/mach3-software/MaCh3/blob/d491feb7a2f59861baf81683292c958ebcacf873/cmake/Templates/Doxyfile.in#L858-L872

KSkwarczynski:
There are some merge conflicts.

Is it possible to use 
`/// `
instead of
```
/**
 *
 */
```
for a doxygen to be consistent with rest of code base?

CLI/Modules/DiagMCMCModule.cpp
alexanderrichards: Whilst the version has to be specified, the `argparse::default_arguments::help` effectively turns it off by not allowing the --version option to be called for this parser.

cli/api/argparse.hpp
alexanderrichards: I don't understand this comment, can you ellaborate?

KSkwarczynski: So mach3 is compiled with very rigorous set compilation flags (Werror, Wall etc.)
Sometimes header files have actual implementation and they "external" headers are being compiled with MaCh3 flags.

Which causes compilation errors.

By doing
```
_MaCh3_Safe_Include_Start_ //{
#include "argparse/argparse.hpp"
_MaCh3_Safe_Include_End_ //}
```
we says compiler to ignore some warnigns/errors for a given header.
Which allows to avoid having annoing errors.

KSkwarczynski:
CI are failing because you need to add argparser etc to pyMaCh3
https://github.com/mach3-software/MaCh3/blob/438172955e7cf2d592c7e537172e1aac44d930b6/cmake/Modules/pyMaCh3.cmake#L80-L94

You should be able to check locally by calling
`RUN pip install --no-cache-dir .
`
https://github.com/mach3-software/MaCh3/blob/438172955e7cf2d592c7e537172e1aac44d930b6/Doc/MaCh3DockerFiles/Alma9/Dockerfile#L49

KSkwarczynski (approved):
LGTM

alexanderrichards:
> CI are failing because you need to add argparser etc to pyMaCh3

So I haven't added these to the pyMaCh3 makefile yet. I'll add that and check it still builds.

alexanderrichards:
Also, one final comment, given that this PR essentially moves the implementation of the three modules so far coded into a new file, just before your ready to merge I will check and pull across the latest implementation from the upstream to make sure we don't miss anything. Just let me know when your ready to merge.

KSkwarczynski:
From my point of view it is mergable. Once you confirm with upstream I think you can merge yourself

KSkwarczynski:
Is it fine to merge?

alexanderrichards:
> Is it fine to merge?

I'm about to merge once this final set of tests passes. I just grabbed the final change that wasn't propagated to the new code so everything should then be fine.

PR #922, opened 2026-06-01
feat: Update CUDA containers

KSkwarczynski:
# Pull request description
New containers have two advantages, first of all newer ROOT version, secondly they are much smaller in memory (because now I know better how to use docker features)

---


henry-wallace-phys (approved):
Approved, with minor question; what is the reason for 2 different root versions (3 if you include the ICPX container)

PR #924, opened 2026-06-01
feat: cmake support for build type

KSkwarczynski:
# Pull request description
Cmake provides several build types
https://cmake.org/cmake/help/latest/variable/CMAKE_BUILD_TYPE.html

We haven't been using this feature at MaCh3, even though it is standard.

The only thing that it does is set MaCh3 options based on Build type.
For example, if build type is debug, then it will enable debugging within MaCh3

The issue has been brought up by Luke

---


henry-wallace-phys (approved):
lgtm

PR #890, opened 2026-05-06
tidy: Remove 7D Loop in splines

KSkwarczynski:
# Pull request description
Long standing issue of binned spline code is 7D loop structure, which make it impossible to extend. As adding any new dimension requires adding new loop...

This PR tackles this issue by using single loop.

For discussion see:
https://t2k-experiment.slack.com/archives/C08MMSCFWDV/p1777973042178879

## Examples

---

- [x] I have read and followed the [contributing guidelines](https://github.com/mach3-software/MaCh3/blob/develop/.github/CONTRIBUTING.md).

PR #928, opened 2026-06-08
feat: implement map for spline index

Charlotte-Knight:
# Pull request description

PR #890 slowed down the spline input/processing. The majority of the slowdown was tracked down to two related functions:
1. `GetSplineIndex` function, which matched `iSample, iOscChan, iSyst, iMode, iVar1, iVar2, iVar3` to a spline by looping through each spline in `IndexVect` and checking for a match, which has O(n) lookup performance. 
2. `isValidSplineIndex` function, which checked that a set of `iSample, iOscChan, iSyst, iMode, iVar1, iVar2, iVar3` corresponded to an entry in `IndexVect`, which worked by looping through every entry in `IndexVect`.

The `GetSplineIndex` function has been replaced by an std::map, and `isValidSplineIndex` has been changed to check for the existence of a key in the std::map. In both cases, the performance improvement is from O(n) -> O(log n). 

I tried using a std::unordered_map but saw no performance benefit.

---

### DIFF
--- a/Splines/BinnedSplineHandler.h
+++ b/Splines/BinnedSplineHandler.h
@@ -111,6 +111,8 @@ class BinnedSplineHandler : public SplineBase {
 
     /// @brief Variables related to determined which modes have splines and which piggy-back of other modes
     std::vector<SplineIndex> IndexVect;
+    /// @brief Map between spline origin/properties (iSample, iOscChan, iSyst, iMode, iVar1, iVar2, iVar3) and the index of the spline in IndexVect
+    std::map<std::tuple<int, int, int, int, int, int, int>, int> IndexVectMap;
 
     std::vector<int > coeffindexvec;
     /// Unique coefficient indices
--- a/Splines/BinnedSplineHandler.cpp
+++ b/Splines/BinnedSplineHandler.cpp
@@ -681,23 +681,7 @@ bool BinnedSplineHandler::isValidSplineIndex(const std::string& SampleTitle, int
 //****************************************
   int iSample = GetSampleIndex(SampleTitle);
 
-  // find matching entry in flat structure
-  bool found = false;
-
-  for (const auto& entry : IndexVect)
-  {
-    if (entry.iSample  == iSample  &&
-       entry.iOscChan == iOscChan &&
-       entry.iSyst    == iSyst    &&
-       entry.iMode    == iMode    &&
-       entry.iVar1    == iVar1    &&
-       entry.iVar2    == iVar2    &&
-       entry.iVar3    == iVar3)
-    {
-      found = true;
-      break;
-    }
-  }
+  bool found = IndexVectMap.find(std::make_tuple(iSample, iOscChan, iSyst, iMode, iVar1, iVar2, iVar3)) != IndexVectMap.end();
 
   if (!found)
   {
--- a/Splines/BinnedSplineHandler.cpp
+++ b/Splines/BinnedSplineHandler.cpp
@@ -46,6 +46,7 @@ void BinnedSplineHandler::CleanUpMemory() {
   //Not a huge saving but it's better than leaving everything up to the compiler
   MACH3LOG_INFO("Cleaning up spline memory");
   CleanVector(IndexVect);
+  IndexVectMap.clear();
   CleanVector(SplineFileParPrefixNames);
   CleanVector(GlobalSystIndex);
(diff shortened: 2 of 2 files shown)


Splines/BinnedSplineHandler.h
```diff
@@ -111,6 +111,8 @@ class BinnedSplineHandler : public SplineBase {
 
     /// @brief Variables related to determined which modes have splines and which piggy-back of other modes
     std::vector<SplineIndex> IndexVect;
+    /// @brief Map between spline origin/properties (iSample, iOscChan, iSyst, iMode, iVar1, iVar2, iVar3) and the index of the spline in IndexVect
+    std::map<std::tuple<int, int, int, int, int, int, int>, int> IndexVectMap;
```
henry-wallace-phys, line 115: Have you checked the lookup times for this vs std::unordered_map?

Charlotte-Knight: Nope, I'll do that now.

dbarrow257, line 115: Is there a way to make this more generic so that it doesn't require 7 variables?

The eventual goal is to move to a system which has generic N dimension spline binning (where N maybe different from 7).

Charlotte-Knight: For MaCh3_DUNE, the EventRates changed from 67 to 66s by using std::unordered map, but it requires you to implement a way to hash the tuple, so the implementation would look like this
```
struct TupleHash {
      std::size_t operator()(const std::tuple<int, int, int, int, int, int, int>& t) const {
          auto hash_combine = [](std::size_t& seed, auto const& v) {
              seed ^= std::hash<std::decay_t<decltype(v)>>{}(v)
                      + 0x9e3779b9 + (seed << 6) + (seed >> 2);
          };

          std::size_t seed = 0;

          std::apply([&](auto const&... elems) {
              (hash_combine(seed, elems), ...);
          }, t);

          return seed;
          }
     };


    /// @brief Map between spline origin/properties (iSample, iOscChan, iSyst, iMode, iVar1, iVar2, iVar3) and the index of the spline in IndexVect
    std::unordered_map<std::tuple<int, int, int, int, int, int, int>, int, TupleHash> IndexVectMap;
```
Which I think is not worth the effort. 

You're right to ask the question though, std::map is only O(log n) whilst std::unordered_map is O(1) which I didn't realise.

Charlotte-Knight:
I made a mistake during my testing of MaCh3_DUNE, meaning that I was building v2.5.0 of MaCh3 core without my additions. And after fixing that, I realised that the spline processing is still slow. I will continue to try to figure out what is causing this. In the meantime, I'll close this PR.

Charlotte-Knight:
I fixed my testing issues from earlier and fixed a further issue with the slowdown (`isValidSplineIndex`). Now, the EventRates app of MaCh3_DUNE runs in 90s. Before PR #890 , it ran in 67s, so there is still a slowdown, but it now completes in a time that is reasonable.

I'll now debug what's going on with the Reweight Validations CI test.

Splines/BinnedSplineHandler.h
henry-wallace-phys: Wow okay that's not worth the improvement

henry-wallace-phys: thanks for checking though!

KSkwarczynski, line 115: Super minor, but can we clean this `IndexVectMap`:
https://github.com/mach3-software/MaCh3/blob/ee0cc57377e92d8e16f241354f102988ffa658a9/Splines/BinnedSplineHandler.cpp#L43-L64

Charlotte-Knight: I think this would be driven by how the generic N-dimensional binning gets implemented in the rest of the code. For example, it could be implemented by flattening the bins across dimensions, and keeping a separate data structure that kept track of what each bin in the flattened structure corresponded to. In that case, the map would only have one int corresponding to the bin number. 

In principle, one can create std::map or std::unordered_map using keys that correspond to custom structures, as long as you can implement something that orders the keys, or hashes the keys. This might be useful if the way the N-dimensional binning is implemented gets more complicated than my suggestion above. So yes, there is a way to make this more generic in future.

KSkwarczynski (approved):
LGTM

PR #936, opened 2026-06-17
feat: Safer Bots [using commit hash]

KSkwarczynski:
# Pull request description
Dan raised, but more is outlined here: https://github.com/github/awesome-copilot/blob/main/instructions/github-actions-ci-cd-best-practices.instructions.md 
that using Hash of action is safer than using tag

---


dbarrow257 (approved):
LGTM

henry-wallace-phys:
If we're doing this for bots, does it also make sense to add this in for the various CPM version in CMakeLists?

PR #940, opened 2026-06-19
breaking: More Wildcards

KSkwarczynski:
# Pull request description
MaCh3 analysis are becoming increasingly more complex.
Wildcards helps to remove desired parameters from plotting etc more easily.

This add wildcard support to 
- LLH Skip vector
- Adaptive Skip
- Process MCMC skip

Previously, they worked based on either string starting with or equality.
Hence, marked as breaking.

## Examples
default
default
after applying Norm_*
exclude

default
before

after applying Norm_*
after

---

### DIFF
--- a/Parameters/ParameterHandlerUtils.h
+++ b/Parameters/ParameterHandlerUtils.h
@@ -42,7 +42,7 @@ namespace M3
 /// @brief Matches a string against a simple wildcard Pattern using regex. Is not case sensitive
 /// @param Text    Input string to test.
 /// @param Pattern Wildcard pattern to match against.
-inline bool RegexMatch(std::string Text, std::string Pattern) {
+inline bool CaseInsentiveMatch(std::string Text, std::string Pattern) {
   // Make a copy and to lower case to not be case sensitive
   std::transform(Text.begin(), Text.end(), Text.begin(), ::tolower);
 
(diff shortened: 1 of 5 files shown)


Parameters/ParameterHandlerUtils.h
```diff
@@ -39,6 +39,27 @@ _MaCh3_Safe_Include_End_ //}
 
 namespace M3
 {
+/// @brief Matches a string against a simple wildcard Pattern using regex. Is not case sensitive
+/// @param Text    Input string to test.
+/// @param Pattern Wildcard pattern to match against.
+inline bool RegexMatch(std::string Text, std::string Pattern) {
```
henry-wallace-phys: it would be nice, I think to also have `RegexMatch(std::string Text, std::vector<std::string> Patterns)` which matches if any entry in the list matches

henry-wallace-phys (approved):
LGTM

Parameters/ParameterHandlerUtils.h
```diff
@@ -39,6 +39,39 @@ _MaCh3_Safe_Include_End_ //}
...
```
luketpickering:
```suggestion
inline bool CaseInsensitiveGlobMatch(std::string Text, std::string Pattern) {
```

luketpickering: or whatever, be precise in function names

Parameters/ParameterHandlerUtils.h
```diff
@@ -39,6 +39,39 @@ _MaCh3_Safe_Include_End_ //}
...
+  catch (const std::regex_error& e) {
+    MACH3LOG_ERROR("Regex error: {}", e.what());
+    return false;
+  }
+}
+
+/// @brief Matches a string against a simple wildcard Pattern using regex. Is not case sensitive
+/// @param Text    Input string to test.
+/// @param Patterns Collection wildcard patterns to match against.
+inline bool RegexMatch(std::string Text, const std::vector<std::string>& Patterns) {
```
luketpickering: also a more correct name here, `CaseInsensitiveGlobAny`?

luketpickering (changes requested):

PR #960, opened 2026-06-30
feat: Umbrella Sampling

DavidRiley12345:
# Pull request description

This PR adds the core features for umbrella sampling. The method works by splitting a single parameter that is difficult to sample, or that has strong degeneracies with another parameter, into several easier-to-sample regions. This helps prevent chains from getting stuck in local regions, allows them to cross likelihood valleys, and improves sampling near the boundaries between degenerate modes. The chains are then combined postfit to reconstruct the global posterior distribution. At present this is limited to `delta_cp`, but the long-term goal is to make the target parameter configurable.

This implementation follows the method described by [Matthews et al.](https://arxiv.org/abs/1712.05024). 
See [Dinner et al.](https://arxiv.org/abs/1705.08445) for many more details. 

### Method Overview

1. The target parameter is split into several easier-to-sample regions. These are called "umbrellas", from the umbrella like shape of an array of overlapping bias functions. 

image

2. MCMC chains are then run independently in each of umbrella region. The chains are kept within their assigned regions by adding LLH penalties according to a user-specified bias function. The output of an umbrella fit is therefore a set of individual separate MCMC chains in each bias region. 
3. These chains are then recombined to reconstruct the global posterior distribution. This is done using the `UmbrellaSolver.C` macro, which calculates the overlap matrix between all of the umbrella sampled chains. The eigenvector of this matrix contains the respective weights of the umbrella windows and can be solved via fixed point iteration of the overlap matrix. 
4. The output of the `UmbrellaSolver.C` macro contains all the steps from the individual chains with their respective umbrella weights saved in a new branch. This output can be plotted with the `UmbrellaPlotting.C` macro, to create 2D contour plots, triangle plots etc.

### Step by step for running a fit

1. Choose a bias function. Current recommendation is Gaussian or Generalised Gaussian depending on how wide a parameter range you are trying to sample and how fine the features are in your posterior. Generalised Gaussian seems to be better at confining chains to parameter regions which are heavily disfavoured in the target distribution. 
2. Choose a bias width. This varies with the rough size of the parameter you are trying to sample. In general this should be smaller than the resolution on the parameter you are trying to sample, else chains drift to the global minima, rather than the bias minima location
3. With a bias chosen you can now select the spacing between the windows. The method works best when neighbouring windows have an appreciable number of steps which overlap. Overlapping your windows at 1-2 sigma seems a good balance between overlap and coverage on the target parameter. This also depends on how long the individual windows will run for, long chains mean more steps in the tails which overlap. On delta_cp ~10 windows equally spaced from -pi to pi has been succesful at current generation sensitivities, and ~60 windows equally spaced is needed for future experiment sensitivities.
4. Current recommendations for other YAML parameters are:

These parameters scale the proposal step size down with the bias window width, recommendation is to enable this feature
 - UmbrellaAdjustStepScale: True 
 - UmbrellaStepScaleFactor: 1.0  
 
These parameters can be used to set the window widths automatically based on the number of equally spaced windows, but hand tuning has been more successful, leave these false
 - AutoOverlapMode: False            
 - UmbrellaNumber: 5 
 - SigmaOverlap: 2 

This is an untested feature to flip step proposals around the middle of the bias window to help sampling at saddle points. Leave false until further testing.
 -  FlipWindow: False 

5. Chains are then run as normal
6. With your chains run you can proceed to use the `UmbrellaSolver.C` macro to process your outputs.

### Usage of UmbrellaSolver.C

1. The macro takes an input YAML file that points to the locations of the individual umbrella chains, or to a folder containing combined parallel chains. It searches for all `.root` files in the folder. If the expected number of root files is found, the fit configuration is extracted from the files and the umbrella bias settings are loaded for use in calculating the overlap matrix. The YAML file also sets the output files, convergence tolerance, print frequency, and maximum number of iterations.
 - You can also manually specify the root file locations along with bin centres and widths, but this is tedious for large numbers of windows and is more prone to entry errors.
7. Files are sorted based on their window centres. This is not required by the method, but it keeps the overlap matrix close to diagonal and makes debugging easier.
8. The macro then performs a precaching step where the umbrella biases are precomputed. This is quite slow and can be very memory intensive, depending on the number of steps and windows required. For around 60 windows and roughly 60 million total steps spread across 24 cores, approximately 32 GB of memory is needed. This step greatly speeds up the matrix iteration phase.
9. After precaching, the fixed-point iteration of the overlap matrix begins. This step is embarrassingly parallel, so it benefits strongly from multithreading. For this to work, the script must be compiled with OpenMP using the compile commands at the top of the macro, or an equivalent setup for your environment.

10. Once the umbrella weights converge, the script can optionally run a robustness check, which randomly perturbs the weights to verify that they are truly stable. This is useful for testing, but it can increase solve time unnecessarily if you are already confident in the convergence.
11. After convergence is reached, the script calculates the bias applied to each event so that the posterior can be reconstructed, and saves everything into a single combined chain containing the events from all umbrella regions.
12. At a configurable rate set in the YAML input, the script also writes out both the overlap matrix and the vector of umbrella weights. This is useful for restarting an umbrella solve if it fails or jobs are cancelled.

### Usage of `UmbrellaPlotting.C`

This is a standard plotting script that reads the output of `UmbrellaSolver.C` and applies the stored weights. You can also write your own plotting script, since all of the umbrella-bias information is stored in the output of `UmbrellaSolver.C`. This script is mainly provided as a starting point.

1. Set the input to the output `.root` file from `UmbrellaSolver.C`.
2. Set the output directory.
3. Optionally enable reference MCMC chains or LLH scans for validation.
4. Run it as a ROOT macro, for example: `root UmbrellaPlotting.C`

## Changes or fixes

- The umbrella method is implemented on top of the existing `MR2T2` algorithm. The algorithm name will have `"_UmbrellaSampled"` appended to indicate that it is not a standard `MR2T2` run. The umbrella penalty is applied in this app after the LLH is calculated from the systematics and before the step is accepted or rejected. Similar additions to other MCMC algorithms would be possible to turn them into umbrella sampled versions.
- `MCMCBase` is updated to store the parameters needed for umbrella sampling, initialise the `MulticanonicalMCMCHandler`, set up multicanonical debug output, and initialise the parameters as needed for umbrella sampling.
- Debug output now includes `MulticanonicalMCMCHandler` messages when enabled.

## Examples

Multimodal posterior in `delm23` versus `delta_cp`. The stripes in the `delta_cp` direction show the unweighted umbrella-biased chains. This demonstrates how a chain can sample across three modes in the distribution, where standard MCMC would get stuck in a single mode.

image

Post reweighting the stripes in delta_cp have disappeared, and we recover two main modes of the distribution across the orderings. The asimov point is well recovered indicated by the red x.

image

Example of umbrella sampled multimodal posterior with refernce delta_cp contour from 3D LLH scan applied to validate posterior shape

image

Sample of an overlap matrix output, illustrating how neighbouring overlap windows show overlap, with the circularity of delta_cp causing the banding structure to wrap at the boundaries of the matrix. The eigen value of this matrix contains the information about the global shape of the distribution.

image

---


Fitters/MulticanonicalMCMCHandler.cpp
```diff
@@ -0,0 +1,345 @@
...
+    MACH3LOG_ERROR("Cannot use multicanonical spline together with umbrella bias function selection.");
+    throw std::runtime_error("Cannot use multicanonical spline together with umbrella bias function selection.");
+  }
+
+  // Umbrella mode requires an explicit bias function selection
+  if (!multicanonicalSpline && !hasBiasFunction) {
+    MACH3LOG_ERROR("Multicanonical umbrella mode requires UmbrellaBiasFunction to be set (gaussian, vonMises, or generalisedGaussian).");
+    throw std::runtime_error("Multicanonical umbrella mode requires UmbrellaBiasFunction to be set.");
+  }
+
```
KSkwarczynski, line 128: Can you use MaCh3Exception?
Like here:
https://github.com/mach3-software/MaCh3/blob/001a3e5effd385ee3f285b96c8acd9f77610c3e1/python/manager.h#L34

MaCh3Exception give line and filename which make debugging bit easier

Diagnostics/UmbrellaSampling/umbrella_solve.yaml
```diff
@@ -0,0 +1,76 @@
...
+    width: 0.04
+    
+  - name: "window_5"
+    center: 0.8726646259971648
+    width: 0.04
+    
+  - name: "window_6"
+    center: 1.0471975511965976
+    width: 0.04
+
```
KSkwarczynski: I would move this config to 
https://github.com/mach3-software/MaCh3Tutorial/blob/main/Tutorial/TutorialDiagConfig.yaml
With field UmbrellPlotting or somethig like this

Diagnostics/UmbrellaSampling/UmbrellaPlotting.C
```diff
@@ -0,0 +1,2007 @@
...
+    int nBins_sin2th12 = 100;
+    int nBins_delm23 = 3000;        // Number of bins for delta_m23^2
+    int nBins_delm12 = 100;
+
+    // Unified plotting ranges requested for all sin2th23 and delm23 views.
+    double sin2th23_min = 0.45;
+    double sin2th23_max = 0.57;
+    double sin2th13_min = 0.015;
+    double sin2th13_max = 0.03;
+    double sin2th12_min = 0.24;
```
KSkwarczynski: Would convert to c++ add to cmake make it compile

Diagnostics/UmbrellaSampling/UmbrellaSolver.C
```diff
@@ -0,0 +1,1283 @@
...
+#include <TLegend.h>
+#include <TList.h>
+#include <TMacro.h>
+#include <TMath.h>
+#include <TObjString.h>
+#include <TStyle.h>
+#include <TSystem.h>
+#include <TSystemDirectory.h>
+#include <TSystemFile.h>
+#include <TTree.h>
```
KSkwarczynski: Would make it c++ and add to cmake to make it compile

Fitters/MCMCBase.h
```diff
@@ -57,6 +57,9 @@ class MCMCBase : public FitterBase {
     /// @param StepsPrint whether to print info about accepted steps and -LogL
     void PrintProgress(const bool StepsPrint = true);
 
+    /// initialise the multicanonical handler
+    std::unique_ptr<MulticanonicalMCMCHandler> multicanonicalHandler = std::make_unique<MulticanonicalMCMCHandler>();
+
```
KSkwarczynski: I would move initialisation to constructor rather than header.

Fitters/MCMCBase.h
```diff
@@ -70,4 +73,13 @@ class MCMCBase : public FitterBase {
...
     double AnnealTemp;
+
+    /// multi-canonical method toggle on/off
+    bool multicanonical;
+    /// multi-canonical penalty
+    double multicanonical_penalty;
+    /// delta_cp parameter value
+    double delta_cp_value;
+    /// dm23 parameter value
+    double delm23_value;
```
KSkwarczynski: I think all these 3 variables are used in single function (ProposeStep).
You can remove them from header and use locally, unles I missed something

Fitters/MR2T2.cpp
```diff
@@ -46,6 +46,30 @@ void MR2T2::ProposeStep() {
...
+#ifdef DEBUG
+    // Print the multicanonical penalty and the delta_cp and delm23 values to the debug file
+    if (debug) debugFile << " delta_cp: " << delta_cp_value << " delm23: " << delm23_value << " multicanonical_penalty: " << multicanonical_penalty << std::endl;
+#endif
+
+        llh += multicanonical_penalty;
+
+    // MACH3LOG_INFO("Delta CP value: {}", delta_cp_value);
+    // MACH3LOG_INFO("Multicanonical penalty: {}", multicanonical_penalty);
+    // MACH3LOG_INFO("LLH after multicanonical penalty: {}", llh);
```
KSkwarczynski: You can make do MACH3LOG_INFO -> MACH3LOG_DEBUG

This way you will have easier way to turn of messages rather than uncommenting.

Fitters/MulticanonicalMCMCHandler.cpp
```diff
@@ -0,0 +1,345 @@
...
+    auto* syst = systematics[static_cast<int>(s)];
+
+    MACH3LOG_INFO("Systematic: {}", syst->GetName());
+
+    // if we find the covariance object
+    if (syst->GetName() == "osc_cov" || systematics.size() == 1) { // hacky, move to the new unified cov obj structure properly
+      oscCovVar = static_cast<int>(s);
+      MACH3LOG_INFO("Found osc_cov systematic saving in variable {}", oscCovVar);
+
```
KSkwarczynski: This is terribly fragile.

What I would do is simply 
```
for(iCov)
   for(iParam)
      if(Param == delta_CP)
           oscCovVar = iCov
```

Still this isn't most robust solution but I think safer

Fitters/MulticanonicalMCMCHandler.cpp
```diff
@@ -0,0 +1,345 @@
...
+  if (!foundDeltaCP) {
+    MACH3LOG_ERROR("Could not find delta_cp parameter in osc_cov systematic");
+    throw std::runtime_error("Could not find delta_cp parameter in osc_cov systematic");
+  }
+  if (!foundDelm23) {
+    MACH3LOG_ERROR("Could not find delm2_23 parameter in osc_cov systematic");
+    throw std::runtime_error("Could not find delm2_23 parameter in osc_cov systematic");
+  }
+
+  return;
```
KSkwarczynski: Purely stylistic, but adding return at the end doesn't do anything, so is considered "noise".

It makes code unnecessarily longer, thus many people discourage it.

Fitters/MulticanonicalMCMCHandler.cpp
```diff
@@ -0,0 +1,345 @@
...
+
+  const auto mcmcConfig = fitMan->raw()["General"]["MCMC"];
+
+  // Get the multicanonical beta value from the configuration file
+  // This acts as a global bias strength factor
+  multicanonicalBeta = mcmcConfig["Multicanonical"]["Beta"].as<double>();
+  MACH3LOG_INFO("Setting multicanonical beta to {}", multicanonicalBeta);
+
+  // TODO DR: I realised this will fail if you set spline to true but don't
+  // delete the umbrella section, thats not ideal
```
KSkwarczynski: you can do /// @todo 
then this will be added here:
https://mach3-software.github.io/MaCh3/todo.html
which make following todo easier I think

Fitters/MulticanonicalMCMCHandler.cpp
```diff
@@ -0,0 +1,345 @@
...
+  // setup for spline bias mode
+  if (multicanonicalSpline) {
+
+    std::string splineFile = GetFromManager<std::string>(mcmcConfig["Multicanonical"]["Spline"]["SplineFile"], "nofile");
+
+    TFile* file = new TFile(splineFile.c_str(), "READ");
+    if (!file || file->IsZombie()) {
+      MACH3LOG_ERROR("Could not open multicanonical spline file: {}", splineFile);
+      throw std::runtime_error("Could not open multicanonical spline file");
+    }
```
KSkwarczynski: if you do TFile *file = M3::Open(splineFile, "READ", __FILE__, __LINE__);
this will cover this if file zombie etc inside making code shorter.

More importantly at the end you should call file->Close();

Fitters/MulticanonicalMCMCHandler.cpp
```diff
@@ -0,0 +1,345 @@
...
+
+void MulticanonicalMCMCHandler::AdjustUmbrellaStepScale(const std::vector<ParameterHandlerBase*>& systematics) {
+  if (umbrellaAdjustStepScale) {
+    MACH3LOG_INFO("Adjusting umbrella step scale to keep ratio of step scale to multicanonical sigma constant");
+    MACH3LOG_INFO("Setting umbrella step scale factor to {}", umbrellaStepScaleFactor);
+    double stepScale = (umbrellaWidth * umbrellaStepScaleFactor) / (2.0 * TMath::Pi());
+    MACH3LOG_INFO("Setting individual step scale for multicanonical separate to {}", stepScale);
+    for (auto& syst : systematics) {
+      if (syst->GetName() == "osc_cov") {
+        syst->SetIndivStepScale(multicanonicalVar, stepScale);
```
KSkwarczynski: Use `oscCovVar `instead of hard-coding "osc_cov"

Fitters/MulticanonicalMCMCHandler.cpp
```diff
@@ -0,0 +1,345 @@
...
+    MACH3LOG_INFO("Not adjusting umbrella step scale, using value in OscCov config");
+  }
+}
+
+void MulticanonicalMCMCHandler::InitializeMulticanonicalParams(std::vector<ParameterHandlerBase*>& systematics) {
+  // Set starting point of chain to umbrella center for non-spline umbrella modes
+  if (!multicanonicalSpline) {
+    for (auto& syst : systematics) {
+      if (syst->GetName() == "osc_cov") {
+        syst->PrintPreFitCurrPropValues();
```
KSkwarczynski: Same as before can you use `oscCovVar` instead of `osc_cov`?

Fitters/MulticanonicalMCMCHandler.cpp
```diff
@@ -0,0 +1,345 @@
...
+    return -(-std::log(dcp_spline_val) + std::log(dcp_spline_NO->Eval(-TMath::Pi() / 2))) * (multicanonicalBeta);
+  }
+  // std::cout << "Evaluating spline at delta_cp = " << deltacp << " gives value
+  // " << dcp_spline_val << "with -log lh of :" << -log(dcp_spline_val) <<
+  // std::endl;
+}
+
+double MulticanonicalMCMCHandler::GetMulticanonicalWeightGaussian(double deltacp) {
+  constexpr double inv_sqrt_2pi = 0.3989422804014337;
+  const double neg_half_sigma_sq = -1 / (2 * umbrellaWidth * umbrellaWidth);
```
KSkwarczynski: Can you write explanation where did you get this number from?

Fitters/MulticanonicalMCMCHandler.h
```diff
@@ -0,0 +1,142 @@
...
+
+/// @brief Helper class for configuring and evaluating multicanonical umbrella weights.
+/// @details this method is used to bias the MCMC sampling of chains to particular regions of the parameter space
+/// The method relies on a bias function defined via a yaml configuration file in experimental specific code
+/// With a bias function defined chains run as normal and are combined after the fact using the information
+/// stored in the chains saved yaml configuration file.
+
+class MulticanonicalMCMCHandler {
+public:
+  enum class BiasFunction { Gaussian, VonMises, GeneralisedGaussian };
```
KSkwarczynski: You can do @cite to have link to bibliography:
https://mach3-software.github.io/MaCh3/citelist.html

You can see an example here:
https://github.com/mach3-software/MaCh3/blob/develop/Doc/MaCh3Web/bibliography.bib
https://github.com/mach3-software/MaCh3/blob/001a3e5effd385ee3f285b96c8acd9f77610c3e1/Fitters/MR2T2.h#L7

PR #961, opened 2026-06-30
bugfix: Change from GetDiagonalError to GetError in python

henry-wallace-phys:
# Pull request description
Getting the error in python used the diagonal error in the covariance matrix. This resulted in numerical instability for parameters like `delm2_12` when large number of parameters are loaded. 

i.e. The nominal of delm2_12 is 0.000018 which was stable when O(80) params are included. When the DUNE flux matrix is used, this changes to  3.167396407145781e-05 which is wildly incorrect.

---


KSkwarczynski (approved):
LGTM

PR #962, opened 2026-07-02
bugfix: Matrix shift now scales with parameters original value

henry-wallace-phys:
# Pull request description
Previously we added a flat 1e-9 to the diagonal of our covariance matrix, now this is 1e-9 * [the nominal error of that parameter].

---

### DIFF
--- a/Parameters/ParameterHandlerUtils.h
+++ b/Parameters/ParameterHandlerUtils.h
@@ -549,7 +549,7 @@ inline bool CanDecomposeMatrix(const TMatrixDSym& matrix) {
 
 // *************************************
 /// @brief Makes sure that matrix is positive-definite by adding a small number to on-diagonal elements
-inline void MakeMatrixPosDef(TMatrixDSym *cov) {
+inline int MakeMatrixPosDef(TMatrixDSym *cov) {
 // *************************************
   //DB Save original warning state and then increase it in this function to suppress 'matrix not positive definite' messages
   //Means we no longer need to overload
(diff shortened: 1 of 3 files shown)


Parameters/ParameterHandlerUtils.h
```diff
@@ -561,16 +561,32 @@ inline void MakeMatrixPosDef(TMatrixDSym *cov) {
...
+  for (int iVar = 0 ; iVar < matrixSize; iVar++) {
+    original_diagonal[iVar] = (*cov)(iVar, iVar);
+  }
 
   for (iAttempt = 0; iAttempt < MaxAttempts; iAttempt++) {
     if (CanDecomposeMatrix(*cov)) {
+      if(iAttempt > 0){
+      	MACH3LOG_WARN("Matrix now positive definite, took {} shifts to fix", iAttempt+1);
+      }
+
```
KSkwarczynski: I feel bad because I think I broke it some time ago...

https://github.com/mach3-software/MaCh3/commit/e15af9c57efa7472a391e07069cccba976afa3d8

You can see originally it had 

```
  if(!use_adaptive || AdaptiveHandler->GetTotalSteps() < 2) {
    MACH3LOG_INFO("Had to shift diagonal {} time(s) to allow the covariance matrix to be decomposed", iAttempt);
  }
```

I think we NEED this printout back.

Maybe if you make MakeMatrixPosDef to return int not void
then you could have this printout with adaptive protection in ParamHandler?

KSkwarczynski (approved):
Otherwise LGTM

Parameters/ParameterHandlerUtils.h
henry-wallace-phys: That makes sense to me

PR #996, opened 2026-07-27
feat: Logarithmic LLH Scan

veeramik:
# Pull request description
Adds a feature to FitterBase that allows users to run an LLH scan using a logarithmic scale, which is especially useful for exotics analyses such as sterile neutrino fits. 

## Changes or fixes
`FitterBase::CalculateBinEdges` calculates logarithmic scale points using the scan ranges defined in a yaml. These are then used in `RunLLHScan` and `Run2DLLHScan` if an option `LLHLogarithmic` is set to true in a yaml. This option is default set to false since I anticipate most people won’t need this feature. If the scan range includes negative values or zero the code just defaults back to a normal LLH scan.

## Examples
[LogLLHExample.pdf](https://github.com/user-attachments/files/30421039/LogLLHExample.pdf)

These are LLH scans made using tutorial for Norm_param_0 (scan range [0.1, 2], generated value 1) and baseline ([1, 300], generated value 295) using both non-log and log scales. 

---

### DIFF
--- a/Fitters/FitterBase.cpp
+++ b/Fitters/FitterBase.cpp
@@ -1,4 +1,7 @@
 #include "FitterBase.h"
+#include "TMath.h"
+#include <vector>
+#include <memory>
 
 _MaCh3_Safe_Include_Start_ //{
 #include "TRandom.h"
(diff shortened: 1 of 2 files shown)


Fitters/FitterBase.cpp
```diff
@@ -958,28 +1003,62 @@ void FitterBase::Run2DLLHScan() {
...
       double central_x, lower_x, upper_x;
       GetParameterScanRange(cov, i, central_x, lower_x, upper_x, n_points, "X");
+      std::vector<double> binEdges_x;
+       
+      //Logarithmic scan 
+      if(LLHLogarithmic){
+
+	if (lower_x < 0.0 || upper_x < 0.0){
+	  MACH3LOG_WARN("Cannot perform logarithmic scan for {} and {} "" with range [{}, {}], [{}, {}], falling back to linear scan", name_x, lower_x, upper_x);
+	}
```
KSkwarczynski, line 1013: Number of arguments don't match hence compilation problem

Fitters/FitterBase.cpp
```diff
@@ -1,4 +1,7 @@
 #include "FitterBase.h"
+#include "TMath.h"
+#include <vector>
```
KSkwarczynski: Can you hide TMath under _MaCh3_Safe_Include_Start_ like here:
https://github.com/mach3-software/MaCh3/blob/ae3fdd75ba71d0b20fcd9b4434b4724a03223d74/Diagnostics/RHat.cpp#L6-L22

We use _MaCh3_Safe_Include_Start_ for any non MaCh3 and non STL include.

Mostly for ROOT to avoid random compilation errors.

KSkwarczynski (approved):

PR #998, opened 2026-07-28
bugfix: prior predictive without any file

KSkwarczynski:
# Pull request description
The issue has been observed by Dan but basically, it should have been possible to run Prior Predictive without passing chain.

For posterior predicive we have to pass file. So when we run predicive we check if posterior file has toys, if it has toys we load them.

Solution is instead of checking just if toys exist we first have to check if posterior file even exist.

---


dbarrow257 (approved):
LGTM

PR #999, opened 2026-07-28
tidy: remove code duplicates and fix typos in umbrella

KSkwarczynski:
# Pull request description
As title says, there was a huge block whether cod is multithreaded or not.
But it is not really needed to have same code written twice.

---

### DIFF
--- a/Diagnostics/UmbrellaSampling/UmbrellaSolver.cpp
+++ b/Diagnostics/UmbrellaSampling/UmbrellaSolver.cpp
@@ -170,8 +170,8 @@ double summedWindowsWeighted(double x, const std::vector<WindowConfig> &windows,
 // Precompute all window evaluations once: cache[i][j][s] = window_j evaluated
 // at samples[i][s]
 // Memory heavy depending on number of steps/cores/windows
-std::vector<std::vector<std::vector<double>>> buildWindowCache(const std::vector<WindowConfig> &windows, const std::vector<std::vector<double>> &samples, bool use_openmp = true) {
-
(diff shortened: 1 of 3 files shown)


Diagnostics/UmbrellaSampling/UmbrellaSolver.cpp
```diff
@@ -985,6 +940,7 @@ void UmbrellaSolver(const std::string &config_file = "umbrella_config.yaml") {
         // Apply random perturbation to z_current
         std::vector<double> z_perturbed = z_current;
         for (size_t i = 0; i < z_perturbed.size(); i++) {
+          /// @todo this may not be thread safe
```
dbarrow257: We should add some large console warning if this hasn't yet been tested to be thread safe

dbarrow257 (approved):
LGTM other than the warning for thread safe

Diagnostics/UmbrellaSampling/UmbrellaSolver.cpp
KSkwarczynski: David been using multithreading problem is `rand()` isn't thread safe
https://stackoverflow.com/a/6161352

Given what David showed it seemed to work I think it is similar issue as we using TRandom for each thread in covarainceBase long time ago.

dbarrow257: Should we fix it before merging then if it's known not thread safe?

KSkwarczynski: I can give it a go

KSkwarczynski: Ok, done

PR #1000, opened 2026-07-29
tidy: Merge RHat calculators

KSkwarczynski:
# Pull request description
Rhat and Rhat high mem were VERY VERY similar to each other.

This PR creates new class Rhat calculator. Goal is to reduce copying pasting between both approaches.
We can't chose one sadly as high mem produces validation that "light" memory can't.
High mem is also significantly slower.

In future we can redcue duplciates even furhter and move to std::vectors. But this is for another PR.
## Examples
[valid.pdf](https://github.com/user-attachments/files/30503678/valid.pdf)

---


dbarrow257 (approved):
LGTM

PR #1023, opened 2026-08-13
feat: add parameter flip groups

Charlotte-Knight:
# Pull request description

The current parameter flip special proposal helps us move between separated regions of high likelihood where it is a single parameter causing a discrepancy. Examples include sin2th_23 and delm2_23 (mass ordering). 

However, if there are correlations with another parameter, then to reach the other region of high likelihood, one needs to flip both parameters simultaneously. See page 11 of https://arxiv.org/pdf/2608.04059. In other words, the exploration of the likelihood can be more efficient if we create a feature where two parameters are flipped at the same time, rather than independently. 

This PR introduces "FlipGroups", which are configured in the yaml files describing your parameters. This option lives under `Systematic:SpecialProposal:FlipGroup`, and here is an example.

```
- Systematic:
    Names:
      FancyName: sin2th_23
      ParameterName: sin2th_23

    SampleNames: ["Tutorial_*"]
    Error: 0.021
    FlatPrior: true
    ParameterBounds: [0, 1]
    # flip octant around point of maximal disappearance (0.5112)
    # this ensures we move to a parameter value which has the same oscillation probability
    SpecialProposal:
        FlipParameter: 0.5112
        FlipGroup: "sin2th_23_sin2th_13"
    ParameterGroup: Osc
    ParameterValues:
      Generated: 0.546
      PreFitValue: 0.546
      PostND: 0.546
    StepScale:
      MCMC: 0.9
    Type: Oscillation

- Systematic:
    Names:
      FancyName: sin2th_13
      ParameterName: sin2th_13

    SampleNames: ["Tutorial_*"]
    Error: 0.0007
    FlatPrior: true
    ParameterBounds: [0, 1]
    SpecialProposal:
        FlipParameter: 0.025
        FlipGroup: "sin2th_23_sin2th_13"
    ParameterGroup: Osc
    ParameterValues:
      Generated: 0.0220
      PreFitValue: 0.0220
      PostND: 0.0220
    StepScale:
      MCMC: 5.0
    Type: Oscillation
```

One can choose whatever name for a group they want; it does not have to include the names of the parameters in that group. Parameters that have not been assigned a group in the config will be assigned a group according to their parameter name.

To implement this, a new `FlipGroup` struct has been added to `ParameterHandlerBase` that contains the indices and flip points for all the parameters in the group. Now, when performing flipping, instead of iterating over all flip parameter indices, we now iterate over groups, and then iterate over indices within the group. And the random number used to decide whether to flip is universal to all parameters in a group.

I tried this in MaCh3Tutorial with the example config shown above. Here is the distribution of the first 20 steps in the chain, illustrating that things are working as expected.

image

---

### DIFF
--- a/Parameters/ParameterHandlerBase.cpp
+++ b/Parameters/ParameterHandlerBase.cpp
@@ -118,14 +118,26 @@ void ParameterHandlerBase::EnableSpecialProposal(const YAML::Node& param, const 
   doSpecialStepProposal = true;
 
   bool CircEnabled = false;
+  std::pair<double, double> circular_bounds;
+
   bool FlipEnabled = false;
+  std::string flip_group;
+  double flip_point;
 
   if (param["CircularBounds"]) {
     CircEnabled = true;
+    circular_bounds = Get<std::pair<double, double>>(param["CircularBounds"], __FILE__, __LINE__);
   }
 
   if (param["FlipParameter"]) {
     FlipEnabled = true;
+    // grab flip group if it exists, otherwise use the parameter name as the group
+    if (param["FlipGroup"]) {
+      flip_group = Get<std::string>(param["FlipGroup"], __FILE__, __LINE__);
+    } else {
+      flip_group = GetParFancyName(Index);
+    }
+    flip_point = Get<double>(param["FlipParameter"], __FILE__, __LINE__);
   }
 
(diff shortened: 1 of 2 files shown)


KSkwarczynski (approved):
LGTM

PR #1034, opened 2026-08-21
feat: GPU Binned spline calclaution

KSkwarczynski:
# Pull request description
Unbinned for a while had ability to use GPU. For binned it wasn't used due to several reasons.

This PR finally add this feature for binned by mostly reusing existing code.

With this PR I also did some renaming and tweaks.
Due to historical reasons Binned and Unbinned converged to very similar structure. But all variables had different names and were minor differences. Final goal is to use `SplineBase `for as much as possible.

## Examples
Doing validations one can see results are almost identical. Bots are complaining because of floating-point precision (GPU uses float only)
Validations

---


henry-wallace-phys:
Would it be possible to get some kind of GPU vs CPU comparison for binned splines?

PR #1043, opened 2026-08-26
bugfix: Add a Start condition for RM

henry-wallace-phys:
# Pull request description
Robbins monro started its update when the Adaptive MCMC matrix starts updating NOT when the throws start. This can result in a rapidly deteriorating acceptance rate. 

Whilst this will eventually recover as the RM scale gets smaller it is inefficient.

---


KSkwarczynski (approved):
LGTM

PR #1056, opened 2026-09-09
feat: Llh Step Scale Optimisation updates

menailj:
# Pull request description
Changes to improve useability of LLH step scale calculation code.
Allow configurability of type of LLH scan used (total or sample).
Option for only certain parameters to be altered, setup using the config file.
Account for correlations with other parameters. Makes sure correlated parameters have the same step scale.

## Changes or fixes
Switch from Sample to Total LLH scan so that optimisation is possible for all parameters, not only those with ND constraint.
Define new configuration element which allows listing of parameter names which should receive a new step scale based on LLH scan.
Read correlation matrix and create an array of correlated variables for each parameter of interest (ie using earlier mentioned list of parameters). For correlated parameters, determine average step scale calculated and apply to all correlated parameters. Does make an assumption of format for correlation matrix.

Also some small changes in header paths so code compiles.

## Examples

---

- [o ] I have read and followed the [contributing guidelines](https://github.com/mach3-software/MaCh3/blob/develop/.github/CONTRIBUTING.md).


Parameters/ParameterHandlerBase.h
```diff
@@ -364,7 +370,10 @@ class ParameterHandlerBase {
   void SetThrowMatrixFromFile(const std::string& matrix_file_name, const std::string& matrix_name, const std::string& means_name);
   /// @brief Perform sanity check to ensure adaption isn't misbehaving before fit starts
   void SanitizeAdaption() const;
+  /// @brief Setup vector of correlations for each parameter
+  void GetCorrelationVector();
```
KSkwarczynski: This isn't used

Parameters/ParameterHandlerBase.h
```diff
@@ -427,8 +436,13 @@ class ParameterHandlerBase {
   std::vector<double> _fUpBound;
   /// Individual step scale used by MCMC algorithm
   std::vector<double> _fIndivStepScale;
+  /// Vector of correlations for given parameter index
+  std::vector<std::map<std::string, double>> Correlations;
+
```
KSkwarczynski: This I don't think is good idea.

We already have CovMatrix

https://github.com/mach3-software/MaCh3/blob/d7c97132f25b2bf672622b657ca44d4965ec3483/Parameters/ParameterHandlerBase.h#L125-L126

I know that is slightly different but it is same information and better to avoid duplicating entities.

Parameters/ParameterHandlerBase.h
```diff
@@ -427,8 +436,13 @@ class ParameterHandlerBase {
...
   /// Whether to apply flat prior or not
   std::vector<bool> _fFlatPrior;
+  /// Defines the group of the parameter
+  std::vector<std::string> _fParameterGroup;
```
KSkwarczynski: Can we avoid moving ParameterGroup to Base for NOW?

There are multiple ParameterHandlers for T2K (beam and ATM):
https://git.hyperk.org/physics/fitters/mach3-hk/-/tree/develop/ParametersT2K?ref_type=heads
This would make this PR "BREAKING".

I think most people would be fine with this change, but I think for consistency we should separate breaking changes from feature PRs.

Is it possible to avoid using groups for now? And we can revisit this in future?

KSkwarczynski (approved):
LGTM

PR #1060, opened 2026-09-10
feat: functional flip

DavidRiley12345:
# Pull request description

This adds the ability to define a parameter flip using a function which contains other parameters in the fit. This has applications for future generation experiments, where for example, delm23_IO is no longer close to -1 * delm23_NO, and we must correct for the differences between the orderings when proposing parameter flips. In the delm23 case this flip can have a different value depending on (among other parameters) delta_cp, giving rise to the need for a functional flip. 

A functional flip is defined in the covariance object for a given parameter, in the special proposals section. You can only chose either a standard flip around a fixed point, or a functional flip. Functional flips can be combined with circular parameters, just like standard flips. The yaml definition must include a list of the parameters to be flipped along with a formula in a string to be parsed into a [ROOT TF1](https://root.cern.ch/doc/v632/classTF1.html). x is used as a placeholder for the variable you are defining, and the other parameters are assigned values from 0 upwards, and must be wrapped in square brackets. 

When the covariance object is parsed, any functional flips are booked into a pending functional flips array, which is then resolved once all parameters have been initialised, allowing you to register flips on values that might come out of order in your yaml.

Functional flipping should be compatible with the new flip groups, and when a functional flip gets resolved after parameter initialisation, it is added to the corresponding flip group to be flipped together. The functional flip will see the state of the parameter before the standard flips in its group are resolved. 

## Changes or fixes

- new FunctionalFlipProposalStruct lives in ParameterHandlerBase

- FlipGroup struct now contains a vector of FunctionalFlipProposals to be evaulated in turn.

- FlipParameterGroup now grabs and stores the proposed parameter values, so that the functional flip can see the pre standard-flip values. If you are planning on chaining together multiple functional flips be careful as this ordering might produce different results than you expect.

## Examples

Example of a functional flip proposal yaml entry

image

Example of fit output during functional flip setup

```
[ParameterHandlerGeneric.cpp:27][info] You specified: 
[ParameterHandlerGeneric.cpp:28][info] Am instead calling the usual non-PCA constructor...
[ParameterHandlerBase.cpp:188][info] Enabling Flipping for parameter sin2th_23 in group sin2th_23 with value 0.5112
[ParameterHandlerBase.cpp:198][info] Enabling Functional Flipping for parameter delm2_23 in group delm2_23 with config Parameters: [delm2_12, sin2th_12, delta_cp, sin2th_13, sin2th_23]
Formula: -x - [0] * ([1] + cos([2]) * sqrt([3]) * tan(asin(sqrt([4]))) * sin(2 * asin(sqrt([1]))))
[ParameterHandlerBase.cpp:162][info] Enabling CircularBounds for parameter delta_cp with range [-3.141592, 3.141592]
[ParameterHandlerBase.cpp:298][info] Enabling FunctionalFlip for parameter delm2_23 with formula '-x - [0] * ([1] + cos([2]) * sqrt([3]) * tan(asin(sqrt([4]))) * sin(2 * asin(sqrt([1]))))'
```

---

- [ x] I have read and followed the [contributing guidelines](https://github.com/mach3-software/MaCh3/blob/develop/.github/CONTRIBUTING.md).


Parameters/ParameterHandlerBase.h
```diff
@@ -365,9 +365,16 @@ class ParameterHandlerBase {
   /// @brief Perform sanity check to ensure adaption isn't misbehaving before fit starts
   void SanitizeAdaption() const;
 
+    struct FunctionalFlipProposal;
+
+
```
KSkwarczynski: This forward declaration is bit awkward.

I know why you did this but it could be a bit confusing

Parameters/ParameterHandlerBase.h
```diff
@@ -376,10 +383,41 @@ class ParameterHandlerBase {
...
+    FunctionalFlipProposal(const FunctionalFlipProposal&) = delete;
+    FunctionalFlipProposal& operator=(const FunctionalFlipProposal&) = delete;
+  };
+
+  struct PendingFunctionalFlipProposal {
+    int target_index = M3::_BAD_INT_;
+    std::string group_name;
+    YAML::Node config;
+  };
+
```
KSkwarczynski: Can you maybe move this into separate header file for example ParameterFlips.h
or something like this, don't have strong opinion about name.

Which maybe is clearer to have all structs related to flip in a single header file.

This is my preference but don't have super strong opinion on this.

KSkwarczynski (approved):
LGTM, only very minor comments

Parameters/ParameterHandlerBase.h
DavidRiley12345: including the Flip groups struct? maybe SpecialProposals.h ?

KSkwarczynski: Yes and Yes

